PR #17, opened 2018-12-06
Fortran interface: add braid_SetPrintFile

beauof:
This commit adds the ability to set the log file name from Fortran code by exposing the C-function in the Fortran interface.

### DIFF
--- a/braid/braid_F90_iface.c
+++ b/braid/braid_F90_iface.c
@@ -1115,7 +1115,13 @@ braid_F90_Name(braid_set_print_level_f90, BRAID_SET_PRINT_LEVEL_F90)(
    return 0;
 }
 
-/* braid_SetPrintFile( ) */
+/* braid_SetPrintFile( )
+ *
+ * User advice: Null-terminate the string fname in your Fortran code, e.g.,
+ * if braidLogFile is of type character(len=128) and braid_spacetime_rank is of type integer:
+ *
+ * write(braidLogFile, "('braid_log.',I0.12,A)") braid_spacetime_rank, char(0)
+*/
 braid_Int
 braid_F90_Name(braid_set_print_file_f90, BRAID_SET_PRINT_FILE_F90)(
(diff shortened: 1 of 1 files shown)


braid/braid_F90_iface.c
```diff
@@ -1115,6 +1115,19 @@ braid_F90_Name(braid_set_print_level_f90, BRAID_SET_PRINT_LEVEL_F90)(
    return 0;
 }
```
jbschroder: @rfalgout  Hey Rob, do we need to do any thing to take a character * from Fortran?  I assume that the below code works on Andreas' local machine, so maybe not.  But before doing this merge, I thought that I'd ping you.

rfalgout: @jbschroder It looks like we do some string passing in a few places in hypre (e.g., HYPRE_IJMatrixPrint), so I guess this should work.  As long as the string name itself (on the fortran side) has the NULL string terminator in it, all should be okay (most of the time; more on this below).  Either Fortran automatically adds the NULL terminator just like in C or the user needs to make sure to do it explicitly.

The reason we originally avoided passing strings around is because Fortran compilers deal with them inconsistently (as I recall).  Some compilers stack all of the string lengths to the end of the argument list for the function while others insert the lengths immediately after the strings themselves.  I assume the former is what happens in most cases today?

@beauof What is the Fortran type you use in the calling code?

rfalgout: I don't think this should cause any problems, so I'm good with merging it.  I had two concerns, but neither should be an issue:

1. Since the string is at the end of the Fortran argument list, it doesn't matter if the Fortran compiler adds the string length to the end of the C argument list or inserts it just after the string.  The result is the same here.

2. Since Andreas is null terminating the string, it will be interpreted correctly in C.

rfalgout: Works for me.

PR #19, opened 2018-12-20
Added GetIter to BraidCoarsenRefStatus of C++ interface

jcc242:


jcc242:
I didn't want both of those commits in this pull request, just the first one ("Added GetIter...") for some reason github thought since I made a pull request future commits to the same branch should be in it!

PR #22, opened 2019-05-09
Run XBraid on multiple time chunks

steffi7574:
This splits up the global time domain into multiple time chunk and runs XBraid's multigrid iterations on each of the time-chunks separately one after the other. After one chunk has finished, the solution at the last time step is used to initialize the next time chunk (involves communication from last to first processor). 

* New user routine: braid_SetNChunks(...) to set the number of chunks that the global domain should be partitioned into (default 1).
* Not yet supported are the following features: 
     - using a shell-vector 
     - user-defined time grid points. 
 A feature check is included that checks if nchunks > 1 and one of the above are active at the same time, throwing an error if so. 

This branch involves some code restructuring:
* Add a loop over the chunks around the main WHILE-loop (the one that performs the multigrid cycling) * Move this main WHILE-loop from braid_Drive() into a subroutine _braid_ChunkDrive(). It's now in _braid.c as it is not a user function.
* Other cycling routines have been moved to _braid.c, as they are not user functions.


rfalgout:
@steffi7574 The new chunk output looks a bit funny in the default case of only 1 chunk.  Just run the examples to see what I mean.  We probably should make some minor modifications here.  Otherwise, this looks pretty good to me.

I found a weird bug that had been lurking for quite some time in the code.  We were including '_util.c' in one of the files (instead of '.h'), and my cosmetic changes to the makefile caused the compile to break.  Took me a bit to see what was going on.  This caused no harm before, but was clearly wrong.

steffi7574:
@rfalgout Are you talking about screen output? What are you referring to?

rfalgout:
@steffi7574 @jbschroder I merged the master into this branch and fixed the conflicts in braid.c.  This should be good to go once we decide to merge the pull request.

jbschroder:
@steffi7574 @rfalgout 
Hi Stefanie, Rob and I wanted to know if you're actively using this time-chunks capability?

steffi7574:
I am not actively using it right now. But I have a braid interface to SU2 that uses it - but it's not under active development right now.

jbschroder:
Thanks for the update!  Rob and I were wondering if this was the way that we want to do time-chunks long-term.  Rob had the good idea of letting the user specify something like a maximum number of time-points allowed, and then the number of chunks could be computed, and you could even work this into FRefine.

So, I guess we're wondering if we want to merge this, or not ... ? What are folks thoughts?

rfalgout:
Let's hold off.  I was going to merge it in earlier this year, but now all of the branches that include it are too out of date to make it worth the effort.  We would be better off introducing it in some other form later (that works with adaptivity, periodicity, etc.).

jbschroder:
That makes sense to me.  Any objections to deleting the pull request ? We could create an issue, with some notes from this discussion, so that we don't forget :-)

rfalgout:
Yes, let's do that.  The time chunk stuff is also integrated into the omgrit branch, which I will have to redo anyway, so making a note for now sounds good and we can revisit it later.

jbschroder:
OK, I'll close this.  See issue 32

PR #23, opened 2019-05-28
Max levels after refinement

jcc242:
This is an initial implementation of setting a maximum number of levels after refinement. It definitely needs a bit more development before merging, namely with the detecting when we have reached maximum refinement. This is currently in a "works for my use case" state.

The new temporal grid hierarchy is initialized at the end of FRefine. At that point in the algorithm, the only criteria we have available to check against is max_refine and tpoints_cutoff. I think it would be ideal to also check after an FCRelax to see if any more refinement was requested by the user, but that would require another InitHierarchy call.

### DIFF
--- a/braid/_braid.c
+++ b/braid/_braid.c
@@ -3213,6 +3213,13 @@ _braid_FRefine(braid_Core   core,
    /* Initialize new hierarchy */
    _braid_CoreElt(core, gupper)  = f_gupper;
    _braid_CoreElt(core, nrefine) += 1;
+   braid_Int newrefine = _braid_CoreElt(core, nrefine);
+   // If we have hit max refinement, then limit the max levels
+   if(!(newrefine < max_refinements))
+   {
+      braid_Int maxLevelAfterRefine = _braid_CoreElt(core, max_levels_after_refine);
+      braid_SetMaxLevels(core, maxLevelAfterRefine);
+   }
    /*braid_SetCFactor(core,  0, cfactor);*/ /* RDF HACKED TEST */
    _braid_InitHierarchy(core, f_grid, 1);
 
--- a/braid/_braid.c
+++ b/braid/_braid.c
@@ -3213,9 +3213,9 @@ _braid_FRefine(braid_Core   core,
    /* Initialize new hierarchy */
    _braid_CoreElt(core, gupper)  = f_gupper;
    _braid_CoreElt(core, nrefine) += 1;
-   braid_Int newrefine = _braid_CoreElt(core, nrefine);
+   braid_Int numrefine = _braid_CoreElt(core, nrefine);
    // If we have hit max refinement, then limit the max levels
-   if(!(newrefine < max_refinements))
+   if(!(numrefine < max_refinements) || (f_gupper >= tpoints_cutoff))
    {
       braid_Int maxLevelAfterRefine = _braid_CoreElt(core, max_levels_after_refine);
       braid_SetMaxLevels(core, maxLevelAfterRefine);
--- a/braid/_braid.c
+++ b/braid/_braid.c
@@ -3529,7 +3529,7 @@ _braid_InitHierarchy(braid_Core    core,
       else
       {
          /* This is the coarsest level */
-         if ( (level > 0) || (!refined) )
+         if ( (level > 0) || (!refined) || (max_levels <= 1) )
          {
             /* If this is a true coarse level (it has a fine grid above it in
              * the current hierarchy) or it is a fine level that was not built
--- a/braid/_braid.h
+++ b/braid/_braid.h
@@ -205,6 +205,7 @@ typedef struct _braid_Core_struct
    braid_Int              io_level;         /**< determines amount of output printed to files (0,1) */
    braid_Int              seq_soln;         /**< boolean, controls if the initial guess is from sequential time stepping*/
    braid_Int              max_levels;       /**< maximum number of temporal grid levels */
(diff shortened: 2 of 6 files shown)


braid/_braid.c
```diff
@@ -3213,6 +3213,13 @@ _braid_FRefine(braid_Core   core,
    /* Initialize new hierarchy */
    _braid_CoreElt(core, gupper)  = f_gupper;
    _braid_CoreElt(core, nrefine) += 1;
+   braid_Int newrefine = _braid_CoreElt(core, nrefine);
+   // If we have hit max refinement, then limit the max levels
+   if(!(newrefine < max_refinements))
```
jbschroder: I would rename newrefine to "numrefine"

jbschroder:
@rfalgout @jcc242 

Hi Joshua,

Let me make sure I understand this feature.  The user gets to set a variable "maxLevelAfterRefine", such that after the maximum number of allowed refinements occurs, the maximum number of Braid levels changes to "maxLevelAfter Refine".  Right?

This lets you generate lots of AMR levels, but then test with a 2-level Braid solver, right?

Overall, your implementation seems reasonable to me, except I left a couple of inline comments on a variable name, and whether we should check w.r.t. tpoints_cutoff.

And, I suppose we should have a regression test for this as well.

What do you think Rob?

jcc242:
Dr. Schroder,

You hit the intention correctly. This is so that we can do FMG and generate some large number of AMR levels through the FMG process, then limit multigrid to a smaller number of levels.

This provides the user interface function braid_SetMaxLevelsAfterRefine so that users can specify what that smaller number of levels is.

I'll think about what kind of regression test we can do for this. The biggest hurdle I still have for this PR is detecting when we refinement is halted due to SetRFactor() all being ones (or not being called?). That seems to be a common scenario in the existing examples/regression tests.

braid/_braid.c
jcc242: Those both sound reasonable, changed here: 5fdcc15c8c6672b09dd258bd0609d7e189efe63d

I mentioned this in my other comment: we probably need to make one other check. How do we detect when our refinement is halted because the step routine no longer calls SetRFactor() but we still haven't met either of the above conditions (max_refinements or tpoints_cutoff)?

jbschroder: I'm not sure how to easily check if the user has stopped calling SetRFactor, either.  You could pick a place in the code, and do an AllReduce on every processors's rfactors.  If the sum is npoints, then the user has stopped calling rfactor (at least for this iteration, it's possible the user could call rfactor the next iteration).

But, Is this something that we really need to check, in that this feature is designed for a very specific use-case?  In other words, is this pull request (minus a little regression test) done?

What do you think Rob? @rfalgout

rfalgout:
I think I need some pictures along with the explanation.  Based on the discussion here, I guess I am wondering why we can't just limit the number of levels throughout the FMG process.  Is this use case namely for debugging?  For example, is it a way to run the exact same FMG refinement cycle that would be generated normally (to produce an identical space-time grid), then have the flexibility to solve on that space-time grid with different cycling strategies?  Or, is there another reason to need this feature (that I am forgetting about, because I'm sure this was explained to me already :)?

jbschroder:
Do you mean like some status or sync hook, where the user can modify
max-levels on the fly with something like
braid_SetStatusMaxLevels(...)
?

rfalgout:
I'm not sure what I mean, because I don't fully understand the goal.  If we want to do a two-level cycle after the FMG refine phase, we could just do two-level cycles all the time.  The code can already do that, and the number of refinements in time and space can be as deep as we like.  If the goal is to do a two-level cycle after the FMG refine phase, but do something non-two-level during it, I can only see a reason for this to debug a new algorithm under development (to ensure we get the same space-time grid each time).  That may well be the answer.  Just asking. :)

jbschroder:
My understanding was the latter goal, to do multilevel FMG refinements, but then use two levels for the post-refinement solve phase (for debugging).  So, that's why I was wondering if what Joshua already has is enough.

rfalgout:
If it is indeed only for debugging, then using the sync hook might be a better option.  Also, for debugging, you don't really need to "detect" when the refinement is finished.  You could first do an unrestricted run, save the number of refinements that were done, then rerun as many times as you like and check against that number.  I'm not sure we have a way to effectively call SetRefine() from within a sync hook, but it should be a safe thing to do there.

jcc242:
Dr. Falgout, I never tried doing the first goal you suggested: setting max levels to 2 (for example) and running with FMG to create multiple AMR levels but keeping the multigrid levels at 2.

This was primarily to achieve the second goal you listed, to help with debugging a new algorithm. Is there any reason for a production code to switch from 5 levels down to 2 levels, for example?

If we want to use a sync hook to set this, then we will need to add another sync call to the program. Currently in _braid_FRefine we first call _braid_InitHierarchy(...) then we call _braid_Sync(...). This is because for my current Sync usage I look at the newly generated hierarchy and put the C-points into my space-time mesh.

rfalgout:
Hi @jcc242 .  I can't think of another reason to do this in a production code or otherwise; only the debugging use case.  We definitely have something left to do with regard to the Sync() idea, but we probably should provide more "synchronization" points in braid anyway.  Also, the issue of calling SetRefine() from within Sync() needs to be sorted out as well.

I will create an issue about this and point to this pull request for more detail, then close this request.  Thanks for taking a closer look!

rfalgout:
Closing this.  See issue #38 .

PR #25, opened 2019-06-24
Step status prior and next C-point

jcc242:
I've added the ability to get the prior and next C-point given a current index. I based this largely off how things are done in _braid_Coarsen/CoarsenRefStatus. Some notes:

1. If you are currently at a C-point before starting your step (if the C-point is the stepStart), then the prior C-point is your current time.
2. If you are on the coarsest grid, both prior and next C-point values are -1. Does XBraid allow negative time values? If so this will need to be changed to some other invalid default (infinity? NaN?)
3. For two calls to _braid_StepStatusInit I wasn't sure what to do. First was in braid_test.c: are there levels and a time grid here? Second is in _braid_base.c where I don't really know what is going on. I added the default/invalid -1 value to there.
4. I'm not entirely sure what is going on with the intermediate coarsened levels when the time domain of the intermediate level doesn't reach the final time points of the finest level. What happens when you try to reach Level 1's last time value + 1? I thought it would segfault for being out of the array's length, but it doesn't. Instead it returns the last C-point of the level, so both the prior and next C-point time values are the same. Is that desired behaviour? I assume the same thing happens in _braid_Coarsen?

Thoughts on how to improve this? Behaviour changes that need to be made?


jbschroder:
@jcc242 @rfalgout 
Hi Joshua,

Regarding your points above,
1) Let, me make sure I unerstand this.  Every time index "k" corresponds to a time value t_k, and that t_k lies in a C-point interval
T_i <= t_k < T_{i+1}
where T_i and T_{i+1} are the time-values for the neighboring C-points.  This pull request then lets the user have access to the time-values T_i and T_{i+1} , even if T_i == t_k.  

2) I think a dummy value of -1 is OK.  Rob?

3) In braid_test.c, there aren't any levels present here.  So, these should likely be dummy values, perhaps [prior, next] = [t, t+dt], instead of [-1, -1].
In _braid_base.c, this is the adjoint capability in braid.  I think that you can ignore this for now.

4) Hmm...let's say that the time-grid on a level looks like
C-F-C-F-C-F'
where the F-block contains "m-1" total time points, except the last F-block -- F' -- contains anywhere from 0 to m-1 time points.  For every time point in that last 
C-F'
block, then the question is what the prior and next C-point time-values values should be.  Right?  I would say that the [prior, next] values should be 
[T_C,  -1]
where T_C is the time-value of that last C-point, and -1 is a marker that you're on the last interval.  Rob

5) Lastly, I suppose we'd need a regression test here.

jcc242:
This is obsolete. The Sync interface function is a much better solution than what I was trying to do here.

PR #27, opened 2019-07-19
Sync interface function

jcc242:
Here is the initial work for a Sync interface function and SyncStatus providing access to XBraid's internals.

1. I'm not sure where all we want to make calls to _braid_Sync. I placed an example one in _braid_FRefine at the location I think I will need it (I might need to move it a bit as I test it more).
2. I added a new AsCaller definition to indicate where in FRefine I am calling Sync.
3. Do we need to do anything for adjoint in _braid_BaseSync?
4. To the core struct I added a function pointer for Sync and a boolean useSync data field. I don't think both are necessary as we could just check if the function pointer is null, but I followed the example set by the optional refinement (though I know you can do temporal refinement without spatial refinement and that is why there are separate checks for function pointer and boolean in that case).
5. I haven't written anything for this in the Fortran interface. I can do that if you'd like. I will likely be expanding the C++ interface in the next couple patches.


jcc242:
Also what do we need to add to the documentation, examples, and tests to support these changes?

braid/braid.h
```diff
@@ -731,6 +741,15 @@ braid_SetSpatialRefine(braid_Core         core,   /**< braid_Core (_braid_Core)
                        braid_PtFcnSRefine srefine /**< function pointer to spatial refinement routine */
                        );
```
jbschroder: This comment before "SetSync" is what get's automatically put into the Users manual.  So, we should make sure that it's good.  Maybe something like,

* Set sync routine with user-defined routine.
* The sync routine gives the user access to XBraid and to the user's app 
* at various points (primarily once per iteration inside FRefine and outside
*  in the main cycle loop).  This function is called once per-processor (not for every 
* state vector stored on that processor, like access).  The envisioned usage here
* is to allow the user to update their app once per iteration based on information
* from XBraid.  This has been useful for time-space adaptivity. 
* Default is no sync routine

braid/braid.h
```diff
@@ -215,6 +215,16 @@ typedef braid_Int
                      braid_AccessStatus  status            /**< can be querried for info like the current XBraid Iteration */
                      );
```
jbschroder: Maybe update the comment a bit to

* The sync routine gives the user access to XBraid and to the user's app 
* at various points (primarily once per iteration inside FRefine and outside
*  in the main cycle loop).  This function is called once per-processor (not for every 
* state vector stored on that processor, like access).

braid/braid_status.h
```diff
@@ -523,6 +554,9 @@ ACCESSOR_HEADER_GET1(Objective, Tol,           Real)
 #define braid_ASCaller_FRefine   2
 /** When CallingFunction equals 0, Braid is in FAccess */
 #define braid_ASCaller_FAccess   3
+/** When CallingFunction equals 4, Braid is inside FRefine after the new finest
+ * level has been initialized */
+#define braid_ASCaller_FRefine_AfterInitHier   4
```
jbschroder: Looks good

jbschroder:
Hi Josua,

I just left some inline comments regarding your questions on
documentation.  For an "advanced" feature like this, I would just rely on
the automatic documentation which just processes comments in .h files.  I
left a couple inline comments in the pull request about updating your
comments.

What I would recommend for examples and tests is to add an option
to ex-01-expanded.c to use the "sync" option.  Perhaps, you could use the
sync function to store a solution value at t-final in the app-structure.
Then, you could add a command line option "-sync" to ex-01-expanded to turn
on sync and print the final solution value to the screen after
braid_Drive() completes.

After that, you'd add a test to braid/test/ode1D.sh to run something like
$RunString -np 1 $example_dir/ex-01-expanded -sync -ml 2
You'd have to update the ode1D.saved file and the "lines_to_check" variable
in the ode1D.sh file to test the new output from sync.  Once you get to
this point, let me know, and we can walk through it if you want.
The whole testing structure is described in braid/tests/README.md, but it's
probably easiest to just walk through a simple example together.

Regarding your other questions,

1) This can be a work in progress

2) Looks good

3) Hmm...I don't think so

4) Let's mimic the spatial refine interface, where we don't use a Boolean.
So, let's get rid of the extra Boolean variable.

5) Supporting this in the Fortran and C++ would be great

I'll read through the rest of the code now, and leave any inline comments
that make sense.

Nice work!  This will be a cool feature.

- Jacob

jbschroder:
Actually, let me correct myself on (3).  I think that you want to "record"
to the actionTape like for other functions.  Or, you'll need to add a check
in _braid_AdjointFeatureCheck() so that no one uses adjoint and sync.

braid/_braid.h
```diff
@@ -278,6 +280,9 @@ typedef struct _braid_Core_struct
    braid_Int     done;             /**< boolean describing whether XBraid has finished */
    braid_Int     wrapper_test;     /**< boolean describing whether this call is only a wrapper test */
    braid_Int     calling_function; /**< from which function are we accessing the vector */
```
jbschroder: Maybe call this tloc_upper and tloc_lower, or something similar (?) to denote that these are local time values?

braid/_braid_status.c
```diff
@@ -382,6 +392,40 @@ ACCESSOR_FUNCTION_GET4(Access, TILD,            Real, Int, Int, Int)
 ACCESSOR_FUNCTION_GET1(Access, WrapperTest,     Int)
 ACCESSOR_FUNCTION_GET1(Access, CallingFunction, Int)
```
jbschroder: Could you add access to tupper and tlower to the Step function as well?  I have a suspicion that this will come in handy later.

jcc242: Added in https://github.com/XBraid/xbraid/pull/27/commits/7c22105086d4a0c86e514ee82db68039c31534b7

braid/braid.h
jcc242: Changed in https://github.com/XBraid/xbraid/pull/27/commits/bd1d1607c288d5e1d073076837313d33585d079c

braid/braid.h
jcc242: Changed in https://github.com/XBraid/xbraid/pull/27/commits/8fc6b99c17f8feb553d0b48a5ee5c9b8c567b981

braid/_braid_status.c
jcc242: Actually its not as simple as in that commit. Every StepStatusInit will need to get the local upper and lower times.

braid/_braid.h
jcc242: Changed in addb4ab5ec2fbb34481dd5f5190a9740ca0aee22

braid/_braid_status.c
jcc242: Looks like its working now, addb4ab5ec2fbb34481dd5f5190a9740ca0aee22

jcc242:
I disabled Sync for Adjoint in 973610b8d5a85979235aa1acd8b278a2f29b5e7b. I think I may need a better understanding of what is needed for Adjoint before adding features for that in.

Changed check for whether to use sync to be similar to scoarsen (i.e. not using a separate boolean) in 3ce3034334e434be751dadd133909a51d0f54f3c.

C++ interface expanded in a76ae63765531cc2e5801c25dbe67404bbee1065.

I need to read the Fortran interface code a bit more to see how it works. 

I'll add the example and test soon.

jcc242:
A basic example and test is in 241b6bbc69907f3565962ec0f97a94879db75221

braid/_braid_status.c
```diff
@@ -148,6 +150,42 @@ braid_StatusGetDone(braid_Status status,
    return _braid_error_flag;
 }
 
+braid_Int
+braid_StatusGetTIUL(braid_Status status,
```
jbschroder: Shouldn't level_ptr be a regular int, and not a pointer?  Its not an input, right, it's just a level index?  Same goes for the next function.

braid/braid.c
```diff
@@ -474,6 +475,7 @@ braid_Drive(braid_Core  core)
    _braid_CycleState  cycle;
    braid_Int          iter, level, done, refined;
```
jbschroder: @jcc242 Can we remove this line?

braid/_braid_status.h
```diff
@@ -103,6 +108,19 @@ _braid_AccessStatusInit(braid_Real          t,                /**< current time
...
+/**
+ * Initialize a braid_SyncStatus structure
+ */
+braid_Int
+_braid_SyncStatusInit(braid_Int           iter,             /**< current iteration in XBraid*/
+                      braid_Int           level,            /**< current level in XBraid */
+                      braid_Int           nrefine,          /**< number of refinements done */
+                      braid_Int           gupper,           /**< global size of the fine grid */
+                      braid_Int           done,             /**< boolean describing whether XBraid has finished */
+                      braid_Int           calling_function, /**< from which function are we accessing the vector */
```
jbschroder: I think you mean "from which function are we accessing the braid" or "..the app"  @jcc242

braid/braid.h
```diff
@@ -731,6 +744,22 @@ braid_SetSpatialRefine(braid_Core         core,   /**< braid_Core (_braid_Core)
                        braid_PtFcnSRefine srefine /**< function pointer to spatial refinement routine */
                        );
 
+/**
+ * Set sync routine with user-defined routine.
+ * Sync gives user access to XBraid and the user's pap at various points
```
jbschroder: "user's pap" to "user's app"  @jcc242

braid/braid_status.h
```diff
@@ -209,6 +219,29 @@ braid_StatusGetDone(braid_Status status,                   /**< structure contai
...
+ * values are returned.
+ **/
+braid_Int
+braid_StatusGetTIUL(braid_Status status,                   /**< structure containing current simulation info */
+                    braid_Int   *iloc_upper,               /**< output, the upper time point index on this processor */
+                    braid_Int   *iloc_lower,               /**< output, the lower time point index on this processor */
+                    braid_Int   *level_ptr                 /**< input, level for the desired indices */
+                    );
+
+/**
```
jbschroder: Is this comment correct?  I think GetTimeValues needs a different doc string here.  @jcc242

jbschroder:
@rfalgout @jcc242 
Hi guys, I just went through the code, and left a few very minor comments.  After that, I'm good to go with this branch.  I verified the new PDF docs and unit test on my machine, too.

braid/_braid_status.c
jcc242: To do this we should change (or add) the ACCESSOR_HEADER_GET macros. Currently they all force a pointer type. I think this would require a new macro that allows mixed input/output:

```
#define ACCESSOR_HEADER_GET3(stype,param,vtype1,vtype2,vtype3) \
braid_Int braid_##stype##StatusGet##param(braid_##stype##Status s, braid_##vtype1 *v1, braid_##vtype2 *v2, braid_##vtype3 v3);
```
where the first 2 arguments are pointers and the 3rd argument is a plain old value. Should I implement this? I'll have to give it a different name than ACCESSOR_HEADER_GET3.

jbschroder:
Hi Joshua,

Oh, I see.  Yes, that makes sense to me.  You could name it something like

ACCESSOR_HEADER_GET3_NOPOINTERS

- Jacob

braid/_braid_status.h
jcc242: Fixed in commit b79b82f58b83a235334f913e70f8e415518ab83c

braid/braid.c
jcc242: Fixed in cbf1bb00ba48443a77aec25da2d16f2a2007f874

braid/braid.h
jcc242: Whoops! Fixed in cbf1bb00ba48443a77aec25da2d16f2a2007f874

braid/braid_status.h
jcc242: After looking at these, I didn't like GetTIUL or GetTimeValues. Changed both in a7eb13f44271569788f222c1f12b16bca28e0002

braid/_braid_status.c
jcc242: I ended up calling it ACCESSOR_HEADER_GET1_IN2 etc. for get 1 value with 2 inputs. See commits 5b12ba731b8748734121727c96a5f51eadc5de54 and b8d6e6f6d7a58c2eb9afe0aee01945aad5832cb8

jcc242:
I think I have addressed all of the comments now. In case it got lost in the shuffle, I created two new macros for mixed input/output. See commits 5b12ba731b8748734121727c96a5f51eadc5de54 and b8d6e6f6d7a58c2eb9afe0aee01945aad5832cb8

Thank you for the feedback Dr. Schroder! Let me know if anything else should be changed.

jbschroder:
Everything looks good, so I'm going to merge. @jcc242 @rfalgout

PR #30, opened 2019-11-18
Interface routine for printing convergence history to a file

steffi7574:
Adding an interface routine (C) that can be called after braid_Drive(), in order to print the convergence history (residuals at all iterations) to a file.

### DIFF
--- a/braid/braid.c
+++ b/braid/braid.c
@@ -1200,32 +1200,52 @@ braid_Int
 braid_FlushConvHistory(braid_Core core,    
                        const char* filename)
 {
-  FILE* braidlog;
-  int niter, ntime, nlevels;
-  int cfactor;
+   braid_Int     myid          = _braid_CoreElt(core, myid_world);
+   braid_Real    tstart        = _braid_CoreElt(core, tstart);
+   braid_Real    tstop         = _braid_CoreElt(core, tstop);
+   braid_Int     gupper        = _braid_CoreElt(core, gupper);
+   braid_Int     nlevels       = _braid_CoreElt(core, nlevels);
+   _braid_Grid **grids         = _braid_CoreElt(core, grids);
 
-  /* Get some general information from the core */
-  ntime = _braid_CoreElt(core, ntime); 
-  braid_GetNLevels(core, &nlevels);
-  _braid_GetCFactor(core, 0, &cfactor);
+   FILE* braidlog;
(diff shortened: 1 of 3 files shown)


jbschroder:
@steffi7574 @rfalgout 
Is this pull request already covered by braid_SetPrintFile() ?  That is, do we need this code?

steffi7574:
As far as I see it, SetPrintFile() redirects standard braid output into the given file.  But braid_FlushConvHistory() prints the residuals ONLY to a file, in a format that is easier to plot since all the string are removed, i.e. no "Braid:" or "|| r_i || =" strings etc. For me it's very useful for plotting e.g. with gnuplot where I just do a "plot <filename> using 1:2". But not sure if you think it's useful for others.

jbschroder:
@rfalgout 
Thanks Stefanie!  I missed that aspect of this pull request.  This seems like a reasonable feature to have.  What do you think Rob, should we merge?

rfalgout:
I am on the fence about this.  It looks like there is quite a bit of shared functionality here.  I see the point about formatting and making it a bit easier to parse, but maybe that argues for changing the formatting in the current print routine to serve both purposes?  We could just merge it and figure out an alternative solution later.  If so, I would change 'Flush' to something else (Write? Save?).  Also, I would prefer that we use '_braid_' versions of the stdio routines (for the same reasons we do it in hypre) and add '_braid_fprintf'.  This could also be changed later.

So, I'm okay with merging this, but I would like to make a few mods at some future time.

jbschroder:
Hmmm...those are good points.  I think the reason why we prepend "Braid:" for the residual history, is so that if Braid is interfaced with another code, it can be easy to find the "Braid" convergence lines among lots of output.  But since we can already redirect Braid output to a special file, I'm not sure how useful that is.  What do you all think?

Regarding using the braid versions of stdio routines, that makes sense to me.  There is a 
_braid_ParFprintfFlush( )
that is like a braid_fprintf, but for parallel, where only rank 0 prints.

jbschroder:
@steffi7574 @rfalgout 
Ok, I changed the name to braid_WriteConvHistory, and use the _braid_ParFprintFlush() instead of fprintf().  I'd say it's ready to merge at this point, if everyone is supportive.

rfalgout:
Works for me.  Thanks!

PR #31, opened 2019-11-18
Allow for non-uniform time-grid refinement

steffi7574:
This allows the user to non-uniformly refine a time-interval by adding time points at prescribed time values. 

Instead of calling `braid_SetRFactor( )` in the users Step routine for uniformly dividing the interval into 'rfactor' subintervals, the user can call 'braid_SetRefinementDTValues(int rfactor, double* dtarray)' that tells xbraid to refine the current time interval based on specified time step sizes, stored in the array 'dtarray'.  

For example, if an interval [tstart, tstop] should be subdivided into 4 non-uniform subintervals, the user passes rfactor=4 and an array of delta-t values of size 3. XBraid's FRefine() routine will add time points at 
t1 = tstart + dtarray[0]
t2 = t1 + dtarray[1]
t3 = t2 + dtarray[2] 

There is currently no test if t3 is smaller than tstop. So the user should be careful when setting up the array of dt values.


braid/_braid.c
```diff
@@ -1428,6 +1429,18 @@ _braid_Step(braid_Core         core,
       rfactors[ii] = _braid_StatusElt(status, rfactor);
       if ( !_braid_CoreElt(core, r_space) && _braid_StatusElt(status, r_space) )
             _braid_CoreElt(core, r_space) = 1;
+
+      /* If NOT called from within FRefine, store the user's dtvalues, if set */
+      braid_Int caller = _braid_CoreElt(core, calling_function);
+      if (caller != braid_ASCaller_FRefine)
```
rfalgout: @steffi7574 This is the only place we use the ASCaller stuff in braid logic (it appears).  Can we do this differently?  Does it really matter where the routine was called?

steffi7574: This was overly complicated, I used it for preventing memory leaks when FRefine call's _braid_Step again. I changed the memory handling, and removed the ASCaller_FRefine restriction.  Now rdtvalues are free'd at the end of FRefine. I added an option to ex-01-refinement to test this feature.

rfalgout:
Hi @steffi7574 .  I modified the approach for setting rfactors and rdtvalues that avoids the memory hoops you had to jump through.  I basically implemented the 'TODO' comment I showed you before.  I tested this with the ex-01-refinement example code (compared it to the version you worked with) and ran that through valgrind as well to ensure that memory is not a problem.

I will need to pull in some other changes from master before I merge this.  I meant to do that first, but forgot.  Let me know what you think.

Thanks!

rfalgout:
Hi @steffi7574 .  Okay, I merged master into this branch.  Had a bunch of issues to sort out for some reason that I don't fully understand.  But, I think this is correct now.  Thanks again!

rfalgout:
@jbschroder Would you mind just taking a look at the new approach I used for implementing braid_StatusSetRFactor?  This pull request is really about adding the variable refinement feature that Stefanie introduced for the power grid project, but I decided to change the way we dealt with rfactor (and hence also rdtvalues) since the previous approach was a result of the fact that status and core were different structures.

You and @steffi7574 both might want to use the "ignore whitespace" setting when reviewing this, since I did some re-indenting and removal of trailing whitespace, etc.

braid/_braid.c
```diff
@@ -3526,10 +3544,14 @@ _braid_InitHierarchy(braid_Core    core,
       /* We need to ensure an rfactor of 1 for global index 0, and for
        * the case of a user-defined residual function and F-relaxation,
        * a default rfactor value at F-points */
-      rfactors[i] = 1; 
+      rfactors[i] = 1;
    }
    _braid_CoreElt(core, rfactors) = rfactors;
 
+   /* Allocate array of refiment dt values, initialized with NULL */
+   rdtvalues = _braid_CTAlloc(braid_Real*, iupper-ilower+2); /* TODO: +2 needed? */
```
jbschroder: From what I can tell, the +2 is probably not needed (?), just ta arrays need the +2, I think.

braid/_braid_status.c
```diff
@@ -352,7 +352,61 @@ braid_StatusSetRFactor(braid_Status status,
...
+   if (level == 0)
+   {
+      _braid_Grid      **grids    = _braid_StatusElt(status, grids);
+      braid_Int         *rfactors = _braid_StatusElt(status, rfactors);
+      braid_Int          index    = _braid_StatusElt(status, idx);
+      braid_Int          ilower   = _braid_GridElt(grids[level], ilower);
+      braid_Int          ii       = index+1 - ilower;
+
+      /* Set refinement factor */
+      rfactors[ii] = rfactor;
```
jbschroder: This is definitely cleaner than the way we used to do it.  Nice Rob! @rfalgout

braid/_braid.c
rfalgout: There is a comment a few lines above this for 'rfactor' that says, "ensures non-NULL".  Leaving the +2 here would do the same and maybe that's a good enough reason.  What do you think?  I could simply change the 'TODO' comment.

jbschroder: Yeah, you could just remove the TODO comment, so as not to confuse people later.  I honestly don't see why the +2 is needed...but someone did earlier (with the non-NULL comment above), so I'm happy leaving the +2.  It's a trivial amount of memory.

jbschroder:
@steffi7574 @rfalgout 
Hi All,
The pull request looks good to me.  This is a much more natural way to SetRFactor, now that status==core.  Also, the variable refinement will be nice to have!

My only comment is whether we want a regression test with ex-01-refinement using this new option (?)  This is something that I can add afterwards, if you want to merge this Rob.

- Jacob

rfalgout:
@steffi7574 Are you okay with the changes?  I'll merge if so.  Thanks!  -Rob

braid/_braid_status.c
```diff
@@ -352,7 +352,61 @@ braid_StatusSetRFactor(braid_Status status,
...
+
+      /* Store dt values */
+      if (rfactor > 1)
+      {
+         braid_Real  **rdtvalues = _braid_StatusElt(status, rdtvalues);
+         braid_Int     j;
+
+         if (rdtvalues[ii] == NULL)
```
steffi7574: Maybe we want to first test if rdtvalues[ii] != NULL, and if so, free the space. This would ensure that rdtvalues[ii] always has the correct length (rfactor-1) for the copy that comes afterwards. We could possibly create a segfault otherwise, e.g. if rdtvalues[ii] has been allocated elsewhere using a different rfactor. Not sure if that would ever happen, but is still somewhat unsafe.

rfalgout: Good point!  Fixed.

PR #33, opened 2019-12-19
Periodic

rfalgout:
Hi @jbschroder ,  I decided to create a pull request now on this branch so we can start the review/discussion process while waiting to see how things turn out with the AMR work.

Some details on the pull request:

- This introduces a periodic capability to braid.  When SetPeriodic() is called by the user, the grid is treated in a toroidal fashion (similarly to what is done in hypre).
- The user specifies a time interval [tstart, tstop] such that tstart and tstop are "the same".
- Braid does not store the periodic tstop time point on each grid level.
- Integration of the last time interval on each level is done by the processor that owns time point tstart, but braid will give the user the time interval values for the last time interval.
- The coarsest grid is solved by FCF relaxation.
- The number of time intervals on the finest level must be evenly divisible by the product of the coarsening factors.
- To do adaptivity, the FRefine() routine will increase the refinement factors to ensure a valid number of time intervals on each new time grid.


braid/_braid.c
```diff
@@ -555,9 +554,15 @@ _braid_AdjointFeatureCheck(braid_Core core)
  * Macros used below
  *----------------------------------------------------------------------------*/
 
+#define _braid_SendIndexOff -2
+#define _braid_RecvIndexOff -2
+
+#define _braid_MapPeriodic(index, npoints)                              \
+   ( index = ((index)+(npoints)) % (npoints) )  /* this also handles negative indexes */
```
jbschroder: @rfalgout 
Isn't 
( K + M ) % M == K % M
?
That is, couldn't we do 
index = index % npoints
?

braid/_braid.c
```diff
@@ -555,9 +554,15 @@ _braid_AdjointFeatureCheck(braid_Core core)
  * Macros used below
  *----------------------------------------------------------------------------*/
 
+#define _braid_SendIndexOff -2
```
jbschroder: I initially thought that "Off" here referred to offest.  We could call this variable
_braid_SendIndexNull
or something like that.  It's a very minor thing...just a thought.

rfalgout: I like that.  I'll make the change.

braid/_braid.c
```diff
@@ -1962,13 +1974,13 @@ _braid_ComputeFullRNorm(braid_Core  core,
          {
             /* Post send to neighbor processor */
             _braid_CommSendInit(core, level, fi, u, &send_handle);
-            _braid_GridElt(grids[level], send_index)  = -1;
+            _braid_GridElt(grids[level], send_index)  = _braid_SendIndexOff;
             _braid_GridElt(grids[level], send_handle) = send_handle;
          }
          _braid_BaseFree(core, app,  r);
       }
       /* Residual from C-point. */
```
jbschroder: @rfalgout Just to make sure that I understand this here -- by using initiali, you will now compute a residual value (possibly nonzero) for the first row of matrix in the periodic case, whereas, for the non-periodic case, we assume the residual is zero.  Right?

rfalgout: Correct.  I am not crazy about the name 'initiali' (it was originally 'initial_ci'), but I couldn't think of anything better.  It kind of represents the location of the initial condition in some sense: in the standard case it's at zero, but in the periodic case it is kind of the rightmost point (which is also at index -1 because of the toroidal thing).  Suggestions welcome or we could just keep it as is.  I'd like to keep the name short-ish if possible.

jbschroder:
Sounds good.  The only other name that came to mind is
initialcond
but, that could also confuse people potentially.  Perhaps the comment on
line 204 in _braid.h, where initiali is introduced, could be expanded a bit
to say that this corresponds to the location of the initial condition.

braid/_braid.c
```diff
@@ -2519,20 +2529,103 @@ _braid_FRefineSpace(braid_Core   core,
...
+      if (size < cfactor)
+      {
+         break;
+      }
+   }
+   if (divisor < npoints)
+   {
+      divisor *= cfactor;
+   }
+   nextra = divisor - npoints;
```
jbschroder: Let's say that npoints = 11, max_levels=2, and cfactor = 2 everywhere.  Then
nextra = divisor - npoints = 2 - 11 = 9.
Right?  That will give us 20 points, and carry about a near uniform refinement.
This is the logic that we're going for, right?  We could alternatively add just 1 point in time, for this case, but that raises the question of where you should add this point.

braid/_braid.c
```diff
@@ -2519,20 +2529,103 @@ _braid_FRefineSpace(braid_Core   core,
...
+ * Helper routine for periodic problems.  Add refinement to ensure the required
+ * number of points.
+ *----------------------------------------------------------------------------*/
+
+braid_Int
+_braid_PeriodicAdjustRFactors(braid_Core   core,
+                              braid_Int   *lnrpoints_ptr,   /* local num refined points */
+                              braid_Int   *gnrpoints_ptr)   /* global num refined points */
+{
+   braid_Int   lnrpoints = *lnrpoints_ptr;
```
jbschroder: This is slick!  I like it.

braid/_braid.c
rfalgout: Actually, you'll get 4-11 = -7, which is messed up.  The logic works if the loop breaks out due to the if statement, but apparently it's not correct when the loop exits because of max_levels.  Let me work on it.  Good catch!

braid/_braid.c
rfalgout: Thanks!

braid/_braid.c
```diff
@@ -2727,9 +2833,27 @@ _braid_FRefine(braid_Core   core,
    {
       _braid_CoreElt(core, r_space) = 0;
    }
-
+      
+   /* If periodic, add refinement to ensure the required number of points */
+   if (periodic)
+   {
+      f_gupper++;
```
jbschroder: Should the incrementing and decrementing of f_gupper go inside the routine?

rfalgout: No.  One is an index and the other is a count (num points).  I probably could reorganize a bit so this incrementing/decrementing isn't done here.  I wasn't crazy about it when I wrote it, but it was easier than introducing a new variable that represents the count.

braid/_braid.c
```diff
@@ -2727,9 +2833,27 @@ _braid_FRefine(braid_Core   core,
...
+      
    /* Compute r_ilower and r_iupper */
-   MPI_Scan(&r_npoints, &r_iupper, 1, braid_MPI_INT, MPI_SUM, comm);
+   {
+      braid_Int  inbuf = r_npoints;
+
+      /* If periodic, adjust the r_iupper index values in the scan by adjusting
+       * the 'inbuf' value on the first time interval */
+      if ( periodic && (ilower == 0))
+      {
```
jbschroder: I don't see why you subtract by rfactors[0] here, but I'm probably just missing something

braid/_braid.c
```diff
@@ -2786,58 +2916,32 @@ _braid_FRefine(braid_Core   core,
...
-      /* Post r_fa send */
-      if (ilower > 0)
+      /* Post r_ta send (to the left) */
+      if ((ilower > 0) || periodic)
       {
-         _braid_GetBlockDistProc((gupper+1), nprocs, (ilower-1), &prevproc);
-         MPI_Isend(send_buf, 2, braid_MPI_REAL, prevproc, 2, comm,
-                   &requests[ncomms++]);
+         _braid_GetBlockDistProc((gupper+1), nprocs, (ilower-1), periodic, &prevproc);
+         MPI_Isend(&r_ta[0], 1, braid_MPI_REAL, prevproc, 2, comm, &requests[ncomms++]);
```
jbschroder: Should we send r_ta[0] or r_ta[-1] ?  I probably have the offsets wrong in my head.

braid/_braid.c
rfalgout: This is a trick of sorts so that the scan will return the correct r_iupper (refined iupper) values.  You need to write down an example to see why it works.  A simple one with two procs should do.  Suppose I am refining the periodic interval by 4, specified at rfactors[0], and suppose the number of refined points (r_npoints) on both processors is 10.  Then, processor 0 should have a refined interval of [-3, 6] and processor 1 should have [7, 16].  I can make this happen by simply decrementing the inbuf value on processor zero by 3 and then do the scan.

braid/_braid.c
```diff
@@ -2924,7 +3041,7 @@ _braid_FRefine(braid_Core   core,
 
    /* Post receives */
    recv_size = f_npoints*(1+isize+1); /* max receive size */
-   recv_buffer = _braid_CTAlloc(braid_Real, recv_size+1);
+   recv_buffer = _braid_CTAlloc(braid_Real, recv_size);
    f_ca = _braid_CTAlloc(braid_Int,  f_npoints);
```
jbschroder: The MPI_Recv below puts (recv_size+1) data elements into recv_buffer.  Don't we want this +1 ?

examples/ex-01-refinement.c
```diff
@@ -107,6 +107,7 @@ my_Step(braid_App        app,
    /* Use backward Euler to propagate solution */
    /* Use forward Euler to estimate the local trucation error */
    (u->value) = 1./(1. + dt)*v;
```
jbschroder: I would argue that this option is a bit more interesting than the zero solution for the periodic case, but I don't have a strong opinion

braid/_braid.c
jbschroder: Sounds good.  The only other name that came to mind is 
initialcond
but, that could also confuse people potentially.  Perhaps the comment on line 204 in _braid.h, where initiali is introduced, could be expanded a bit to say that this corresponds to the location of the initial condition.

braid/_braid.c
jbschroder: Ok, thanks for the clarification

braid/_braid.c
jbschroder: Ahh...I see.  Thanks!

braid/_braid.c
rfalgout: This is correct.  I am send my left time value to the processor to my left (which will go in a ghost region on that processor).

jbschroder:
@rfalgout 
Cool!  This will be a great feature to have, Rob.  Thank you!  

I just finished going through the code (man, FRefine is a fun function!).  I think that there's still one or questions left there, mostly from me not knowing FRefine very well.

A few final comments are 
- We should insert a regression test or two, to test a few basic test cases
- The user's manual should be updated to advertise this feature, and perhaps explain things a bit, like with that picture you sent a few days ago when I was confused about the block structure of the space-time matrix.  
- The manual should mention how you set the number of relaxations on the coarsest periodic grid (set nrelax for level==maxlevels-1)
- Do we have a check so that SetSeqSoln and Periodic are not combined?

Other than that, I think this is a great new feature!

braid/_braid.c
rfalgout: Actually, the MPI_Recv only needs size, not (1+size), because size already accounts for the extra message-length data I prepend to each send buffer.  I suppose that could cause an implementation of MPI to do something illegal with the buffer memory and cause a memory problem, even though there should be no actual messages received that would overrun the recv_buffer.  I'll try making that change and see if I am correct here.

braid/_braid.c
rfalgout: I added the word 'condition'  :)

braid/_braid.c
rfalgout: I'm pretty sure I have the logic right now.

examples/ex-01-refinement.c
rfalgout: I'll add this for the periodic case then.

braid/_braid.c
rfalgout: Changed (1+size) to size in the MPI_Recv.  It works for me, valgrind turned on.

rfalgout:
Hi @jbschroder .  I agree on your comments, but would like to keep the load balance work as the top priority for now.  Once this is all done, I can address some of what you say.  Feel free to help out with them if you feel comfortable and have time. :)

FYI, I do have a check somewhere about SetSeqSoln, but this should be looked at carefully (I didn't really worry about it).

braid/_braid.c
```diff
@@ -2519,20 +2529,109 @@ _braid_FRefineSpace(braid_Core   core,
...
+   braid_Int  divisor, nextra, level, cfactor;
+
+   /* Compute coarsening factor divisor */
+   divisor = 1;
+   nextra  = 0;
+   for (level = 0; level < (_braid_CoreElt(core, max_levels) - 1); level++)
+   {
+      _braid_GetCFactor(core, level, &cfactor);
+      divisor *= cfactor;
+      if ((int)(npoints/divisor) < cfactor)
```
jbschroder: Should the cfactor in this if-statement, be the cfactor from level+1?

braid/_braid.c
```diff
@@ -2519,20 +2529,109 @@ _braid_FRefineSpace(braid_Core   core,
...
+      {
+         break;
+      }
+   }
+   if ((npoints % divisor) > 0)
+   {
+      /* Product of coarsening factors (divisor) does not evenly divide npoints */
+
+      /* If we can coarsen one more time, update the divisor to reflect that */
+      if (level < (_braid_CoreElt(core, max_levels) - 1))
```
jbschroder: I'm not sure why this if-statement is here?  If this if statement evaluates to true, then that means the above break statement was executed.  Doesn't that mean that we can't coarsen any more?

rfalgout: If we are in this if statement, then the product of the coarsening factors does not evenly divide npoints, so we have to do something.  The previous loop exits in one of two ways: 1) we get to a grid size that is too small (i.e., the break statement); 2) we hit max_levels.  This if statement makes sure we didn't hit max_levels.  If we didn't, then when we increase npoints, we will be able to add an additional grid level and the divisor will be larger than currently computed, hence we increase it by the last cfactor we considered in the loop.

braid/_braid.c
rfalgout: You are correct.  Good catch!  I need to put the if statement first and adjust the logic a tad.  This will also take care of the case where npoints is less than cfactor to begin with.

braid/_braid.c
jbschroder: OK, that makes sense.  Thanks for explaining it!

PR #36, opened 2020-01-15
Changed number of levels to check in PeriodicCheckNumPoints to also consider min_coarse

jcc242:
Previously _braid_PeriodicCheckNumPoints assumed that the simulation would always be using max_levels. This fix also takes into account the min_coarse parameter, allowing for the simulation to have fewer levels than the max.

### DIFF
--- a/braid/_braid.c
+++ b/braid/_braid.c
@@ -2536,6 +2536,39 @@ _braid_FRefineSpace(braid_Core   core,
 }
 
 /*----------------------------------------------------------------------------
+ * Helper routine for periodic problems.  Compute the max number of levels
+ * from user settings
+ *----------------------------------------------------------------------------*/
+
+braid_Int
+_braid_PeriodicNumLevels(braid_Core   core,
+                         braid_Int    npoints,
+                         braid_Int   *nlevels_ptr)   /* nlevels */
+{
+   braid_Int min_coarse = _braid_CoreElt(core, min_coarse);
+   braid_Int max_levels = _braid_CoreElt(core, max_levels);
+
+   braid_Int level, cfactor;
+   braid_Int levelPoints = npoints;
+   braid_Int max_coarse_level = max_levels;
+
+   for (level = 0; level < max_levels; level++)
+   {
+      _braid_GetCFactor(core, level, &cfactor);
+      if((levelPoints/cfactor) < min_coarse)
+      {
+         max_coarse_level = level;
+         break;
+      }
+      levelPoints = (levelPoints/cfactor);
+   }
+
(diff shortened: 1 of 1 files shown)


braid/_braid.c
```diff
@@ -2535,6 +2535,39 @@ _braid_FRefineSpace(braid_Core   core,
...
+   braid_Int max_levels = _braid_CoreElt(core, max_levels);
+
+   braid_Int level, cfactor;
+   braid_Int levelPoints = npoints;
+   braid_Int max_coarse_level = max_levels;
+
+   for (level = 0; level < max_levels; level++)
+   {
+      _braid_GetCFactor(core, level, &cfactor);
+      if((levelPoints/cfactor) < min_coarse)
```
jbschroder: I think that this is " <= min_coarse"

braid/_braid.c
```diff
@@ -2535,6 +2535,39 @@ _braid_FRefineSpace(braid_Core   core,
    return _braid_error_flag;
 }
 
+/*----------------------------------------------------------------------------
+ * Helper routine for periodic problems.  Compute the max number of levels
+ * from user settings
```
jbschroder: Perhaps this would be a better comment?
* Helper routine for periodic problems
* Computes the total number of times that Braid can coarsen, given npoints total time-points,
* and the min_coarse, max_levels, and cfactors values stored in core

And could function title be something like 
_braid_ComputeNumCoarsenings_Helper() ?
I feel like that may be more informative

rfalgout:
This looks much more complex than I thought would be required and it appears to involve quite a bit of code replication.  Why is the fix not a simple change to the original _braid_PeriodicCheckNumPoints routine (one that just adds the additional check on min_coarse in the inner if statement)?

braid/_braid.c
```diff
@@ -2546,12 +2579,14 @@ _braid_PeriodicCheckNumPoints(braid_Core   core,
...
    /* Compute coarsening factor divisor */
    divisor = 1;
    nextra  = 0;
-   for (level = 0; level < (_braid_CoreElt(core, max_levels) - 1); level++)
+
+   _braid_PeriodicNumLevels(core, npoints, &computedmax);
+   for (level = 0; level < computedmax; level++)
    {
       _braid_GetCFactor(core, level, &cfactor);
       if ((int)(npoints/divisor) < cfactor)
```
jbschroder: @rfalgout Do you mean an additional check here like
&& (int) (npoints/divisor) < min_coarse
break;

jcc242:
Looks like this can be closed since it is replaced with https://github.com/XBraid/xbraid/pull/37

rfalgout:
This issue is being addressed in #37 instead.  Closing this one.

PR #37, opened 2020-01-20
Adding a check for min_coarse in PeriodicCheckNumPoints

rfalgout:
Hi @jcc242 and @jbschroder .  Please take a look at this approach to incorporating min_coarse into the periodic stuff.

Joshua: Would you please test against your code to make sure this works for you?

Jacob: I just noticed that the periodic regression tests need to be modified in a couple of ways.  First, we need to specify the number of times points to be a power of two.  So, we should add '-nt 64' to each of the current tests.  Second, we should run this on a wider range of processor combinations (currently we just run on 2, and the code hangs with 1 or 4).  So, doing '-np 1' and '-np 4' for example would be good.  Actually, another good run to do would be to have more procs than intial time points, so for example, '-np 8' and '-nt 4'.  Feel free to make the updates directly on this branch.  (I really need to dive into the regression test stuff so I can contribute more; just being lazy again for now :)


jbschroder:
Hi Rob,

I just pushed some new regression tests to this branch.  It tests the
eventualities that you mention.
 - But I didn't see the code hang for np=1 or np=4
 - And, some runs still take maxiter iterations, and essentially stagnate

Otherwise, this pull request looks good to me!

- Jacob

braid/_braid.c
```diff
@@ -2546,15 +2546,17 @@ _braid_PeriodicCheckNumPoints(braid_Core   core,
...
+   braid_Int  min_coarse = _braid_CoreElt(core, min_coarse);
    braid_Int  divisor, nextra, level, cfactor;
 
    /* Compute coarsening factor divisor */
    divisor = 1;
    nextra  = 0;
-   for (level = 0; level < (_braid_CoreElt(core, max_levels) - 1); level++)
+   for (level = 0; level < (max_levels - 1); level++)
    {
       _braid_GetCFactor(core, level, &cfactor);
```
jbschroder: This check based on min_coarse looks correct to me, but Joshua should test his code on it, too.

jcc242:
I pulled in this change and ran all my tests. Everything looks good from my end, no regressions. I am getting the correct number of time points in my periodic case. I think we can merge this in.

rfalgout:
Thanks for checking on this @jcc242 !

Issue #4, opened 2018-07-11
Splitting up large braid.c and util.c files


jbschroder:
Here's my opinion about naming conventions...

I personally like separating user interface from internal interface.  User's include `.h` files and call functions, so they should only include the `braid.h` file and call functions that begin with `braid_`.  Because of this, we need `.h` files that begin with `_braid_` to put all other header information that users should not see.  However, the `.c` files really can have any names we like and implement any of the functions, either internal or external or both.  In other words, I'm not sure we need to name any of the `.c` files with prefixes `braid_` or `_braid_`.

I would still vote for putting user interface implementations in `.c` files that begin with `braid_` just to make them easy to find, but I am okay with also putting internal functions in these files when it makes sense (helper functions, for example; `braid.c` does this).  So, I would prefer `_braid_status.c` to be `braid_status.c`.  I don't think we need to name any of the `.c` files with `_braid_`.  So, I would be in favor of splitting and renaming the `_braid.c` and `_util.c` files into a number of `.c` files (that do not begin with underscore) that better reflect the code.  For example, `_braid.c` could be split into files such as

  distribution.c
  communication.c
  uvector.c
  step.c
  residual.c
  interp.c
  restrict.c
  refine.c
  space.c
  hierarchy.c
  ...

Maybe we shouldn't split it up quite so much?  Just some thoughts to ponder.

-Rob

rfalgout:
This issue was addressed in pull request #39

PR #39, opened 2020-01-25
Split _braid.c into smaller files

rfalgout:
This is a first version that splits the implementation into smaller files as outlined in issue #4 .  The following doesn't need to be done right now, but should be addressed at some point:

- Further work should be done regarding the organization of include files.
- Further work needs to be done with the file naming convention.


jbschroder:
This looks great. Thank you!

A couple questions that come to mind
- The new files like "norm.c" aren't called "_braid_norm.c", even though they generally have "_braid" functions in them.  This seems inconsistent with our naming of files so far, but does lead to shorter, easier to read filenames.  What do we want to do here?

- Do we want a README file in braid/braid that describes a bit about how the *.c and *.h files are named and how they interact?   I think that this could be good for us, and for any new developers.
     - It looks like the header files are largely untouched, and the new files generally define functions from braid.c and _braid.c, for user-interface and braid-internal functions, respectively.
     - On a related note, the old source files have little preambles for doxygen, with things like

        /** \file _util.c
           * \brief Source code for utility routines. See _util.h for more documentation.
        */
       
        Do we want little headers like this in the new files?  Just a line or two describing what they do?


- I noticed that the "LLNL-CODE-#######" is dropped from the preamble.  I don't particularly care either way, but thought it was worth mentioning in case we need it.

-  _braid_GetUInit is used in residual.c and step.c, but is located in step.c.  Should we move it to uvector.c?

Possible file name changes:
- distribution.c   or   par_distribution.c ?

rfalgout:
Hi @jbschroder .  Thanks for looking this over!  Here are some responses to your comments:

- Regarding file names, issue #4 argues for having user header files and implementation files to begin with 'braid'.  This makes it easy to keep track of what users see.  Otherwise, file names can be anything and should not start with an underscore.  There is possibly one exception with '_braid.h' (but maybe we should call this something else?).  This is a "catch all" include file that developers use to get access to everything, and users could also include it at their own risk.  I personally like simplifying the user header file question by giving them only one that they need to worry about, 'braid.h'.  We currently have a few others floating around.

  So, when I said we had a few more things to organize, here is a more specific list of the files and what I think we should additionally change (and we could do it now if you agree):

  access.c
_braid_base.c           --> base.c
_braid_base.h           --> base.h
_braid.c                --> not sure about this one; migrate to other files over time?
braid.c
braid_defs.h
braid_F90_iface.c
_braid.h
braid.h
braid.hpp
braid_status.c
_braid_status.h         --> status.h
braid_status.h
_braid_tape.c           --> tape.c
_braid_tape.h           --> tape.h
braid_test.c
braid_test.h
communication.c
distribution.c
drive.c
grid.c
hierarchy.c
interp.c
Makefile
mpistubs.c
mpistubs.h
norm.c
refine.c
relax.c
residual.c
restrict.c
space.c
step.c
_util.c                 --> util.c
_util.h                 --> util.h
uvector.c

- I like the idea of adding a README.

- I can't comment on the doxygen question until I understand what this does.  In hypre, we only put doxygen stuff in the user header files, and that is used to create the reference material.

- When I change the license, the LLNL-CODE number will only be in README.md, so we can leave it out of the individual files.  I realized we were missing a few headers, but didn't like the names being there since that list of developers has grown.  I didn't change all of them, though.

- The _braid_GetUInit function is specific to step and residual.  If anything, we might rename it, but it is not a 'braid_UGet...' routine that belongs in uvector.c.

Let me know what you think about these points and we can go from there.  Thanks again!

jbschroder:
OK.  I should have re-read Issue 4 before commenting -- sorry!

Overall, what you suggest makes sense to me, Rob.  A couple small notes

1.  _braid.h/c  could probably be moved into a new adjoint.h/c with other functions going places like hierarchy, and util.  What do you think?  Should we make this switch now?  I worry that if we wait, we'll forget about it.

2. Regarding doxygen, yes, only the header files should have a little documentation at the top.  I suppose with the new header files, we should make sure they all have some.

3.  And yes regarding the README, we should do that.

So...I suppose the next question is splitting this work up and then merging?  Since I've been managing the doxygen stuff, I can take (2).  And I'm happy to do (1) or (3), as well, assuming you like the idea of (1).

rfalgout:
Hi @jbschroder .  Let me take a crack at 1 and 3 and then hand it off to you.  I’ll do that today.  Thanks!

rfalgout:
Okay, @jbschroder .  1 and 3 are done.  Your turn. :)

jbschroder:
OK!  Sounds good.  I will start chipping away this.  It shouldn't take long.

- Jacob

jbschroder:
OK, I'm happy with this.  Thanks for leading the charge on the file split effort Rob!  I suppose we can close issue #4 too.

- Jacob

rfalgout:
Okay @jbschroder .  I'll take care of the merge and issue #4.  Thanks for your help on this!

Issue #41, opened 2020-06-24
Change default access level in example 4


Torrencem:
There's a python script for visualizing the output of ex-04, but unlike ex-02, it doesn't by default output the relevant .out files. This is because the default access level is set to 0 instead of 1. This is a really small change that hopefully makes this example less confusing.

Torrencem:
PR solves the issue (#42)

PR #42, opened 2020-06-24
update and fix example 4's default access level

Torrencem:
Per #41


jbschroder:
These changes look good to me, and make the default access level consistent with other drivers, e.g., ex-02

Issue #43, opened 2020-06-24
Fix visualization scripts to work with either python 2 or python 3


Torrencem:
The visualization scripts both don't support python 3, but with a few future imports they work on either.

Torrencem:
PR solves the issue

PR #44, opened 2020-06-24
add support for python 3 in example visualizers

Torrencem:
#43


jbschroder:
Hi Matt,  Thank you for these pull requests!  I probably won't get to these
until next week, but don't think that I've forgotten about them.

Best,

Jacob

PR #49, opened 2020-07-13
Add Cython Interface and example drivers

jbschroder:


jbschroder:
Adding new Cython interface in braid.pyx, three new Cython examples, and one Cython driver, along with regression tests.

PR #50, opened 2020-07-23
Richardson

jbschroder:
Here is Ben's old Richardson code, updated and with some example capabilities.  It took a while!  But, it's a cool feature that lingered in this branch for too long.


braid/access.c
```diff
@@ -68,7 +68,7 @@ _braid_FAccess(braid_Core     core,
 
          if (access_level >= 1)
          {
-            _braid_AccessStatusInit( ta[fi-ilower], fi, rnorm, iter, level, nrefine, gupper,
+            _braid_AccessStatusInit( ta[fi-ilower], fi-1, rnorm, iter, level, nrefine, gupper,
```
rfalgout, line 71: @jbschroder Has this been a bug all along?  Can you explain what's behind this change.  Thanks!

braid/braid.c
```diff
@@ -858,7 +877,14 @@ braid_SetCRelaxWt(braid_Core  core,
                   braid_Int   level,
                   braid_Real  Cwt)
 {
-   braid_Real  *CWts = _braid_CoreElt(core, CWts);
+   braid_Real  *CWts     = _braid_CoreElt(core, CWts);
+   braid_Int richardson  = _braid_CoreElt(core, richardson);
+
+   if( richardson && (Cwt != 1.0) )
+   { 
+      printf("Weighted relaxation and Richardson extrapolation are incompatible.  Turning off weighted relaxation.\n");
```
rfalgout, line 885: @jbschroder What's our policy again on 'printf' usage?

braid/braid.c
```diff
@@ -1499,3 +1525,48 @@ braid_Rand(void)
...
+   /* Compatability checks */
+   if ( local_order < 2 && ( est_error || richardson ) )
+   {
+      _braid_printf(" Local Order of Time Integrator must be > 1 \n\n" );
+      abort();
+   }
+
+   if(CWt_default != 1.0)
+   {
+      printf("\nWeighted relaxation and Richardson extrapolation are incompatible.  Ignoring Richardson options.\n\n");
```
rfalgout, line 1552: @jbschroder Looks like we have a `_braid_printf` that we should be using?

braid/access.c
jbschroder: @rfalgout Good question!  I found that StepStatus always returns the time-index minus one, but that AccessStatus just returned the time-index.  So, I modified AccessStatus to be the same as StepStatus.  

Overall, I don't quite like the minus one idea, but it has gotten woven into things a bit with StepStatus (and possibly in user codes, too), so I hesitated to change that.  We could put this in the issue tracker as something to resolve.  Let me know what you think!

braid/braid.c
jbschroder: @rfalgout Good point!  I'll change that to _braid_printf here in a minute.  So many coding changes...and it must have slipped my mind :-)

braid/access.c
rfalgout: Isn't there a bigger risk of impacting user code if we change to -1?  Users already call Access, correct?  I agree that we might want to discuss via an issue.  I sort of think that the current behavior makes sense, though.  Basically, status corresponds to the "current" time point, and in the context of `Step` that is `fi-1` because it takes that value and moves it forward.  In the context of `Access` the "current" time point is `fi`.

jbschroder: I was thinking that user code for Access is mostly for outputting things to the screen or file, so the chance for impact would be small.  But I see your point about the time-index.  Inside of Step, tindex is basically the index of "tstart", but inside of Access, we can think of tindex as just the current time-point.  Is that correct?

If we want to keep it that way, I can roll back this change to AccessStatus and make a note in the comments/docs.  Let me know what you think.

rfalgout: Correct.  The term "current" clarifies both situations, I think, because in Step point `fi` is the "next" time point (aligned with `tstop`).  I think we should roll these back.

braid/braid_status.c
```diff
@@ -359,8 +359,13 @@ braid_StatusSetRFactor(braid_Status status,
...
       braid_Int          ii       = index+1 - ilower;
 
-      /* Set refinement factor */
-      rfactors[ii] = rfactor;
+      /* Set refinement factor
+       * 
+       * Note: make sure index ii is positive.  There is one step with
+       * Richardson in FCRelax that is with a C-point from the left-neighboring
+       * processor.  Thus, ii can be negative. */
+      if( ii >= 0 )
```
rfalgout, line 367: @jbschroder I think I will need an explanation for this one as well.  I didn't immediately see how or why `FCRelax` causes this problem.  I am also wondering how this (and the rest of the Richardson stuff) plays with the periodic feature.  Thanks again!

braid/relax.c
```diff
@@ -36,18 +36,88 @@ _braid_FCRelax(braid_Core  core,
...
    _braid_Grid   **grids    = _braid_CoreElt(core, grids);
    braid_Int       ncpoints = _braid_GridElt(grids[level], ncpoints);
-
+   
    braid_BaseVector  u, u_old;
    braid_Real        CWt;
    braid_Int         flo, fhi, fi, ci;
    braid_Int         nu, nrelax, interval;
 
+   /* Required for Richardson */
```
rfalgout, line 45: @jbschroder I would like to think about how we might reduce the impact of Richardson on readability of this routine (and maybe some others), but that can wait until a later date, after we merge this pull request.

jbschroder: @rfalgout Yes, I had some similar thoughts.  But, I frankly started to run out of steam on this PR.  We could try writing some helper functions for relax.c and restrict.c, perhaps.

braid/braid_status.c
jbschroder: Richardson needs to take a step with the C-point to the left of the left-most F-point (on processor).  And sometimes this C-point needs to come from the neighboring processor to the left.  In that case, ii can be negative, and the current processor doesn't "own" this point.  So, I figured that any rfactor requests on this point should be ignored.

Regarding periodic, that's a good point.  I didn't test that.  Is there a case with periodic where ii can be negative, but not the initial condition, and we might want to refine?

braid/braid.c
rfalgout: FYI, note that there are other places that use `printf` .  I can change them if you want.  Just let me know.

braid/braid_status.c
rfalgout: Okay, I see where this happens in `FCRelax` (I think).  However, I don't think users should be setting rfactor for any of these cases, since it's based on the coarse step size.  Does this ever break the code, i.e., cause an out-of-bounds array issue or something?  In looking at this `braid_status.c` code, I'm not sure `ii` is ever negative since it's set as follows:

    braid_Int          ii       = index+1 - ilower;

Basically, we store the rfactors using the "right" index for the interval.

braid/access.c
jbschroder: Sounds good.  I just pushed some changes to address this.  One issue is that when accessing the Richardson error estimate array, Step() needs it's time index shifted by one, whereas Access does not.  This led to the creation of two different StatusGetSingleErrorEst functions, one for Step and one for Access.  But, this was the best solution that I saw.

rfalgout: Looks good.  Thanks!

braid/braid_status.c
jbschroder: We can chat on the phone, if that would be easier.  Just let me know. 

This extra Step() for Richardson happens in relax.c from lines 191 -- 214.  The corresponding communication (if needed, for this C-point corresponding to the left-most interval) is done above in lines 91 -- 120.

And, I agree that users should not be setting RFactor in this case.  One possible fix would be to set "level" to "level+1" in lines 201 -- 202 in FCRelax, so that the StepStatus structure is initialized for a coarse level.  This way, even if the user set rfactor for this step, it would be ignored, because (level == 1)  

Aside: I was getting an array out of bounds, because the time index being stepped from (line 201 relax.c) is (ci - cfactor - 1).  So even (index+1 - ilower) could be negative if cfactor was > 2.

jbschroder:
No problem! Thanks for catching these things.

braid/braid_status.c
jbschroder: So, the proposed changes would be to set "level" to "level+1" in lines 201--202 in FCRelax, and to then also remove the check in SetRFactor "if( ii >= 0 )"

jbschroder: This could be a little confusing to a user, if they are using a different spatial grid or time-stepper (or anything else) on the coarser grid, because Step does pass a "fine-level" vector to the user here.  But, this is a weird algorithmic step.  It's with a fine-level vector, but with a coarse-level dt.

rfalgout: We will leave as is for now and create an issue to make it easier for users to know when it's okay to set rfactor.

rfalgout:
The periodic changes look great, too, @jbschroder !  Thanks!

jbschroder:
Cool!  I'll merge here in a little bit.

PR #56, opened 2020-08-06
Fix bug when _braid_BaseStep is passed a null ustop

Torrencem:
When you pass ustop == null into _braid_BaseStep, there's a null derefence (because of ustop->userVector). This should probably just pass NULL to core->step as ustop, since this is a pretty valid case. This PR is an example of the fix, it doesn't change any behavior besides fixing the segfault when ustop is null.


rfalgout:
Hi @Torrencem .  We are not supposed to call BaseStep with `ustop = NULL`, because we don't ever want to pass a NULL value for this to the user's Step routine.  We do use a NULL value for ustop in `_braid_Step()`, but line 55 of `step.c` helps to ensure that the user gets an appropriate value.  There are a couple of other places where BaseStep is called, and they appear to be okay also.  If not, then we should fix them.

Torrencem:
Ok, that sounds reasonable. The problem came about when I was trying to use _braid_Residual in my code, internally it called _braid_GetUInit which returned a null value, so then _braid_BaseStep crashed. We just emailed you the background to why we were doing this. Thanks for the response

PR #65, opened 2020-12-29
Cython adjoint

jbschroder:
This PR is meant to bring in as many changes as possible from Eric's XBraid fork.  Ideally, we want TorchBraid to be able to compile with Braid Master (if possible).


braid/_braid.h
```diff
@@ -222,6 +222,8 @@ typedef struct
    braid_BaseVector  *va_alloc;      /**< original memory allocation for va */
    braid_BaseVector  *fa_alloc;      /**< original memory allocation for fa */
 
+   braid_BaseVector   ulast;         /**< stores vector at last time step */
```
jbschroder, line 225: @rfalgout @steffi7574 @eric-c-cyr 
TorchBraid assumes that braid core always stores the last time point (whether F or C) in this variable ulast.  

Right now, ulast is only updated in Access and FCRelax when done==1, i.e., after the cycling. But this does move storage questions about vectors out of the Get/Set routines and into parts of the main code.  I'm not sure how I feel about this.  

We could we get rid of ulast inside of braid, and instead tell the user via status whether done==1, and whether this is the last time point.  Then, the user could always store ulast in app.  I'm not sure if this does everything Stefanie and Eric need in Torchbraid. 

Another alternative would be to just always store to ulast in the Get/Set routines, so that the extra logic for ulast is not needed in FCRelax or Access.  This does create a little more work during a Braid cycle, though.

Or, we could leave it as it is.

braid/drive.c
```diff
@@ -648,13 +648,39 @@ _braid_Drive(braid_Core  core,
       _braid_FRestrict(core, level);
    }
 
-   /* Allow final access to Braid by carrying out an F-relax to generate points */
-   /* Record it only if sequential time stepping */
-   if (max_levels > 1)
+
+   /* Allow final access to Braid by carrying out a relaxation to generate points */
+   if ( _braid_CoreElt(core, finalFCrelax) )
```
jbschroder, line 653: @rfalgout @steffi7574 @eric-c-cyr 
Here is the reworked logic about when to call the final FCRelax or the final FAccess.  Right now, only FCRelax OR FAccess is called at the end of a Braid run.  This saves work.

But this does mean that FCRelax needs to do everything FAccess does, and so I've added User-Access hooks into FCRelax, including at time t==0, which needed some extra logic.

A couple more notes, if max_levels is 1, we now rely on FCRelax to do the serial solution in time.  Eric and Stefanie may want to double check this works with everything on their end.  For instance, both Access and FCRelax should now store ulast if done==1.  The unit tests in TorchBraid pass, at least.  But double checking that everything works for max_levels==1 would be good.

braid/relax.c
```diff
@@ -60,17 +72,65 @@ _braid_FCRelax(braid_Core  core,
...
+            {
+               if (_braid_GridElt(grids[level], ulast) != NULL)
+               {
+                  _braid_BaseFree(core, app, _braid_GridElt(grids[level], ulast));
+                  _braid_GridElt(grids[level], ulast) = NULL;
+               }
+               _braid_BaseClone(core, app,  u, &(_braid_GridElt(grids[level], ulast)));
+            }
+         }
+         else if( (level == 0) && (ci == _braid_CoreElt(core, initiali)) && done)
```
jbschroder, line 127: @rfalgout @steffi7574

Here, you can see one of the new User-Access hooks FCRelax needed.   This is the hook to make sure access is given at time t==0, if Braid is done.  The other new User-Access hooks are above.

braid/_braid.h
rfalgout: Maybe we should just always store `ulast` and basically think of it as an extra C-point.  Then Get/Set only need to worry about it.  Do Get/Set have access to `done`?

braid/uvector.c
```diff
@@ -269,6 +312,21 @@ _braid_USetVector(braid_Core        core,
       _braid_BaseFree(core, app,  u);              /* free the vector */
    }
 
+   /* If braid is finished, make sure the last time point is stored, i.e.,
```
jbschroder, line 315: @rfalgout @eric-c-cyr @steffi7574 
Hi All -- I just moved the storage logic for ulast into the USetVector routine below.  Per the discussion with Rob, I think this is a better place for it.  All the regression tests for Torchbraid past, so I think everything  is OK.  But feel free to double check, too.

braid/uvector.c
```diff
@@ -201,6 +201,45 @@ _braid_UGetVector(braid_Core         core,
...
+   braid_Int           gupper   = _braid_CoreElt(core, gupper);
+   braid_Int           done     = _braid_CoreElt(core, done);
+   braid_BaseVector    ulast;
+   braid_Int           num_procs, myid;
+
+   MPI_Comm_size( comm, &num_procs);
+   MPI_Comm_rank( comm, &myid );
+   
+   /* If Done, get vector at last time step*/
+   if(done && (myid == (num_procs-1) ) )
```
jbschroder, line 224: @rfalgout @steffi7574 
As you can see, I'm only making ulast available if "done" is true.  USetVector only stores into ulast if done is true.  If this will conflict with any desired functionality, let me know.

rfalgout (approved):
Looks good to me, @jbschroder .  Thanks!

PR #70, opened 2021-01-22
Send time information to neighboring empty processors for AMR subcycling purposes

jcc242:
When constructing a new hierarchy, this sends time information to empty processors. Empty processors are defined as those that have no time points on them. 

Time information on empty processors is needed for AMR subcycling. We need the empty processors to have the correct next and previous times so that during Sync we update AMR information such as the previous and next support level time point. This information is needed on empty processors so that if the processor later becomes used (like from a temporal refinement), then we have this information ready. 

Essentially the way this works is:
1. If this is NOT an empty processor:
1.1. Send our lower time value (ta[0]) to empty processors to the left. There is an assumption that time processor 0 is never empty.
1.2. Send our upper time value (ta[iupper-ilower]) to empty processors to the right.

2. If this is an empty processor:
2.1 Post receives to the first non-empty processor to the left and right
2.2 Fill the ta[-1] time value with the value received from the left processor
2.3 Fill the ta[0] time value with the value received from the right processor

### DIFF
--- a/braid/hierarchy.c
+++ b/braid/hierarchy.c
@@ -42,6 +42,7 @@ _braid_InitHierarchy(braid_Core    core,
    braid_Int     *rfactors   = _braid_CoreElt(core, rfactors);
    braid_Real   **rdtvalues  = _braid_CoreElt(core, rdtvalues);
    braid_Int      nlevels    = _braid_CoreElt(core, nlevels);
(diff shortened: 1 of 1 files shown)


jbschroder:
Hi All, I'm going to close this now, but we can re-open it, if needed, no problem.

PR #72, opened 2021-02-18
Relaxation only coarsegrid

jbschroder:
Add an option to solve the coarsest grid with relaxation only (braid_SetRelaxOnlyCG).
This is not the typical way to solve a coarsest grid, and is only recommended in specific circumstances.

### DIFF
--- a/braid/hierarchy.c
+++ b/braid/hierarchy.c
@@ -173,8 +173,17 @@ _braid_InitHierarchy(braid_Core    core,
       }
       else
       {
-         /* This is the coarsest level */
-         if ( (level > 0) || (!refined) )
+         
+         /* If solving coarsest grid by relaxation, initialize a standard coarse grid*/
+         if( 1 )
+         {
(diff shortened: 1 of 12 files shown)


braid/hierarchy.c
```diff
@@ -173,8 +174,17 @@ _braid_InitHierarchy(braid_Core    core,
       }
       else
       {
-         /* This is the coarsest level */
-         if ( (level > 0) || (!refined) )
+         
+         /* If solving coarsest grid by relaxation, initialize a standard coarse grid*/
+         if( relax_only_cg && (gclower < gcupper) )
```
jbschroder, line 179: @rfalgout Hi Rob, This is the spot in hierarchy.c that I tweaked to create a normal C/F splitting on the coarsest grid.  Take a look at the logic, if you have a moment, to double check.

rfalgout (approved):
Looks great, @jbschroder !

PR #73, opened 2021-02-18
Feature sync get proc

jcc242:
Here are a few functions I added to allow access to the temporal communicator during the Sync function. I also added a call to the user's Sync function after performing initialization. I found that I required a sync here in order to exchange spatial grids when doing a restart.


jcc242:
Any chance this can get merged in so I can use upstream XBraid? Thanks!

braid/braid_status.h
```diff
@@ -579,6 +596,16 @@ ACCESSOR_HEADER_GET1(Sync, CallingFunction,  Int)
 ACCESSOR_HEADER_GET1(Sync, NumErrorEst,      Int)
 ACCESSOR_HEADER_GET1(Sync, AllErrorEst,      Real)
```
jbschroder, line 598: @jcc242 
I guess the same question is here, can we use the ACCESSOR_HEADER style to declare SyncStatusGetTComm ? I've never tried this for a non-built-in type like MPI_Comm, but its worth exploring, I think.

braid/braid_status.h
```diff
@@ -654,20 +681,22 @@ ACCESSOR_HEADER_GET1(Objective, Tol,           Real)
...
 /** When CallingFunction equals 4, Braid is inside FRefine after the new finest
  * level has been initialized */
-#define braid_ASCaller_FRefine_AfterInitHier   4
+#define braid_ASCaller_FRefine_AfterInitHier 4
 /** When CallingFunction equals 5, Braid is at the top of the cycle */
-#define braid_ASCaller_Drive_TopCycle   5
+#define braid_ASCaller_Drive_TopCycle 5
 /** When CallingFunction equals 6, Braid is in FCrelax */
 #define braid_ASCaller_FCRelax 6
+/** When CallingFunction equals 6, Braid just finished initialization */
```
jbschroder, line 698: This should be "... equals 7, Braid just finished initialization */ "
Thanks for fixing those typos above!

braid/braid_status.h
jcc242: @jbschroder We can make it use the ACCESSOR_HEADER and similarly for the implementation. I specifically chose not to because I wanted to guard access to the temporal communicator. That is, I didn't want to make it easy for someone to add access to temporal communicator to the other status functions and structs.

If you think it would be better to be consistent (or think guarding the temporal communicator in this manner is unnecessary), then I will use the ACESSOR_HEADER/FUNCTION methods.

jbschroder: @rfalgout What do you think Rob?  I probably have a slight preference for consistency, i.e., using the ACCESSOR_HEADER/FUNCTION approach -- but I'm flexible, too.

rfalgout: Hi @jcc242 and @jbschroder .  Sorry for the delayed response.  Isn't the goal of Sync to give the user access to the temporal communicator to do (essentially arbitrary) work in a temporally-synchronized manner?  If that is indeed the goal, then I think it's perfectly fine to give them access to the communicator through a Sync accessor function/macro, and that would be consistent with how we are currently doing things elsewhere (which admittedly isn't perfect).  So, I would vote for consistency as well, assuming I understand the issue correctly here.  We can prevent access in other status functions via the PR process.  Thanks!

jcc242: @rfalgout @jbschroder Hi! I have updated everything to fit your comments. I know use the ACCESSOR_HEADER/FUNCTION approach for GetTComm. I did have to create a new macro for this: ACCESSOR_FUNCTION_GET1_GENERIC

The reason for this new macro is that the standard ACCESSOR_FUNCTION_GET1 prepends the argument type with "braid_", but we don't have a braid_MPI_Comm type. This new macro does not prepend the argument types. Do you think I should `typedef MPI_Comm braid_MPI_Comm` instead of creating this new macro?

Thank you!

rfalgout: I think you should add the typedef in `braid_defs.h` near the other MPI-related stuff.  Thanks!

jcc242: @rfalgout I have done so now.

braid/braid_status.h
```diff
@@ -390,6 +393,18 @@ braid_StatusGetRNorms(braid_Status status,                 /**< structure contai
                       braid_Real  *rnorms_ptr              /**< output, XBraid residual norm history, of length *nrequest_ptr* */
                       );
 
+/**
+ * Returns the processor number in *proc_ptr* on which the time step *index*
+ * lives for the given *level*.  Returns -1 if *index* is out of range.
+ * This is used especially by the _braid_SyncStatus functionality
+ **/
+braid_Int
```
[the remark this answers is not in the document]
jcc242: @jbschroder Yes, I have added that now.

jbschroder (approved):
@rfalgout  This PR looks good to me Rob.  What do you think?

rfalgout (approved):
Looks good to me.  Thanks!

PR #84, opened 2022-11-01
Braid timer

jbschroder:
@rfalgout @eric-c-cyr  
Hi Guys,
Here is a PR that incorporates 
(i) some basic timers (I keep re-implementing these for various projects, so I tried something more permanent here)
(ii) the user-defined MPI buffer that we need for GPU->GPU direct communication

Let me know what ya'll think!


rfalgout:
Hi @jbschroder .  Thanks for the PR!  These will be nice additions to the library.  As we discussed briefly today, I think it would make sense to allow the timing stuff to be turned on/off by defining a `_braid_MPI_Wtime()` that just returns 0 when timing is turned off and otherwise calls `MPI_Wtime()`.  This is not a big deal, though, and it could always wait until later.  Thanks again!

rfalgout (approved):
Looks good.  Thanks!

jbschroder (commented):
Looks good to me

PR #85, opened 2022-11-17
Add tinfo to bufstatus

jbschroder:


jbschroder:
This info (time index and level) is needed in BufferStatus for the GPU work that Eric and Jens are doing.  

The old way to specify things like level and time index inside of BufPack() and BufUnpack() was to insert such information into the message itself, like a header.  But with GPUs, the message is received on the GPU, and they want this info on the CPU.  So, this helps them avoid transfers.

Anyway...Take a look Rob, and let me know what you think!

rfalgout (approved):
Looks good, Jacob!

jbschroder:
@rfalgout Thanks Rob!  If you have any ideas on how to refactor the timings, let me know.  I'll probably merge tomorrow or so.  

In other news, the new user allocate buffer stuff is working nicely in the Torchbraid code.  Jens has gotten some XBraid-GPU scalings to work on his cluster in Wuppertal, which is cool!

PR #86, opened 2023-02-08
Braid delta correction

cadaverous-lives:
@jbschroder , @rfalgout,
Implements the low rank Delta correction, including a new example and a couple of drivers. Currently passing all regression tests on my machine!
(Full Documentation still needed)
Excited to do my first pull request!
-David


jbschroder:
Hey @cadaverous-lives , there are three .pyc files included in the PR.  I think these can be removed.

.gitignore
```diff
@@ -0,0 +1,32 @@
+# output files
```
jbschroder, line 1: Is this a new file? 
@rfalgout @cadaverous-lives 
Do we want to distribute a default gitignore file?  I'm a little on the fence here.  Could this interfere with someone's existing .gitignore?

braid/_braid.h
```diff
@@ -213,6 +236,7 @@ typedef struct
    braid_Real        *ta;            /**< time values                (all points) */
    braid_BaseVector  *va;            /**< restricted unknown vectors (all points, NULL on level 0) */
    braid_BaseVector  *fa;            /**< rhs vectors f              (all points, NULL on level 0) */
+   braid_Basis       *ba;            /**< time-dependent bases       (all points, NULL if not delta_correct or on level 0) */
```
jbschroder, line 239: Isn't the braid_Basis already included in the braid_BaseVector ua, va, and fa?  Why do we need a separate declaration?  Is it regarding the two different bases above?  If so, we should probably specify which time-dependent basis this refers to.

.gitignore
cadaverous-lives: I didn't really intend to leave my gitignore in there, sorry! I was just using it to simplify my own usage.

rfalgout: We do include a `.gitignore` file in hypre now: https://github.com/hypre-space/hypre/blob/master/.gitignore.  It's a good way to prevent certain files from accidentally being committed, and convenient for developers.  It does not override a global `.gitignore`, it adds to it.  If we do include it here, we should at least discuss its contents.  That's my 2 cents.

braid/_braid.h
```diff
@@ -132,16 +135,36 @@ struct _braid_VectorBar_struct
 }; 
 typedef struct _braid_VectorBar_struct *braid_VectorBar;
 
+/**
+ * This contains two arrays of @ref braid_Vector objects which should be thought of
```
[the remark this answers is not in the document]
cadaverous-lives: The wording here is outdated. Only one array of braid_Vectors is stored in each BaseVector. I thought I had updated this, but I guess I forgot to! I've got this fixed now on my local copy.

braid/_braid.h
```diff
@@ -132,16 +135,36 @@ struct _braid_VectorBar_struct
...
- * information.  Stores both the user's primal vector (braid_Vector type) and
+ * information.  Stores the user's primal vector (braid_Vector type) and
  * the associated bar vector (braid_VectorBar type) if the adjoint
- * functionality is being used.  If adjoint is not being used, bar==NULL. 
+ * functionality is being used, as well as basis vectors used by Delta correction.
+ * If adjoint is not being used, bar==NULL, and if Delta correction is not used,
+ * basis == basis_action == NULL.
+ * 
+ * For Delta correction, *basis* stores the unmodified basis vectors, while basis_action
+ * stores the
```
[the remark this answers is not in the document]
cadaverous-lives: See above ^

braid/_braid.h
cadaverous-lives: So, this is a compromise made to save on communication costs. If we wanted to avoid having to store ba, we would need some extra communication on the coarse grid. The problem is that the Delta correction needs to know the restricted fine-grid basis at the previous time-point, so that would be stored in va[i-1], which might be owned by a different processor. Currently, communication on the coarse grid only involves sending ua[-1], so having to send both ua[i-1] and the basis from va[i-1] would nearly double the communication. Maybe there is a smarter way to do this, since I am realizing that va and ba store the same data, just offset by one time-index.

Question: I know that during frestrict, on the coarse grid, the previous time-point is sent to the current processor if it is not owned, but is this value stored in va after frestrict is finished? i.e. if I tried to access va[i-1] from the step function on the coarse grid, would that work?

rfalgout: It looks like `va[-1]` is updated here: https://github.com/XBraid/xbraid/blob/master/braid/restrict.c#:~:text=/*%20Initialize%20update,%7D

So, what you want may already be there.

cadaverous-lives: So this is weird, I figured I would try using the basis at va[i-1], since on one processor it should be equivalent, but for some reason it is failing to converge now. I will have to take some time to figure this out, but it looks like it should be possible to eliminate the ba variable.

cadaverous-lives: I think I've figured this out. Using va[i-1]->basis in place of ba[i] works perfectly up until fInterp, since fInterp uses va to store the error corrections, va[i-1] can be overwritten. I guess I will have to just find a way around this, but it seems doable. I'll work on this more later today.

.gitignore
jbschroder: I could see including a .gitignore being beneficial.  Often, my git status is swamped with temporary files.  

David: I think if you can add a couple comments explaining the contents, including the file isn't a bad thing.

braid/_braid.h
jbschroder: Sounds good...saving this communication step would be nice!

braid/_braid.h
```diff
@@ -320,6 +347,16 @@ typedef struct _braid_Core_struct
    braid_Real             localtime;        /**< local wall time for braid_Drive() */
    braid_Real             globaltime;       /**< global wall time for braid_Drive() */
 
+   /** Delta correction and Lyapunov vector estimation*/
+   braid_Int              delta_correct;    /**< turns on Delta correction to potentially accelerate convergence */
+   braid_Int              delta_rank;       /**< for low rank Delta correction */
+   braid_Int              delta_defer_lvl;  /**< Delta correction will be turned off until this coarse level */
```
jbschroder, line 353: Does delta_defer_lvl  only run delta correction on the coarsest k-levels, or is it the other way around, delta correction is only run on the finest k-levels?

braid/_braid.h
```diff
@@ -369,10 +406,12 @@ typedef struct _braid_Core_struct
    braid_Real    old_fine_tolx;    /**< Allows for storing the previously used fine tolerance from GetSpatialAccuracy */
    braid_Int     tight_fine_tolx;  /**< Boolean, indicating whether the tightest fine tolx has been used, condition for halting */
    braid_Int     rfactor;          /**< if set by user, allows for subdivision of this interval for better time accuracy */
+   braid_Basis   lvectors;         /**< if Delta correction is set, contains reference to the tangent vectors to be propagated across the c-interval */
```
jbschroder, line 409: maybe this should be declared up above with ua, va, and fa?  Seems maybe more relevant there?

braid/_braid.h
rfalgout: How persistent do you need the ba[] array to be?  I thought the va[] array was potentially destroyed after interpolation (it is destroyed in an FMG cycle so I don't think the code touches it again after interpolation).  If you need a similar persistence with ba[] then maybe there is a way to restructure things so you get a valid value from va before it is overwritten (the overwriting is just to save on memory and reduce memory management overhead).

braid/access.c
```diff
@@ -95,7 +95,7 @@ _braid_FAccess(braid_Core     core,
          if ( access_level >= 1 )
          {
             _braid_AccessStatusInit( ta[ci-ilower], ci, rnorm, iter, level, nrefine, gupper,
-                                     done, 0, braid_ASCaller_FAccess, astatus);
+                                     done, 0, braid_ASCaller_FAccess, u->basis, astatus);
```
jbschroder, line 98: I'm guessing this is how you give the user access to the basis at this point?

braid/base.c
```diff
@@ -321,6 +412,41 @@ _braid_BaseSum(braid_Core        core,
    _braid_CoreFcn(core, sum)(app, alpha, x->userVector, beta, y->userVector);
    _braid_CoreElt(core, timer_user_sum) += _braid_MPI_Wtime(core) - timer;
 
+   /* Sum over the basis vectors */
+   if ( x->basis && y->basis )
+   {  /* Add together the bases if they both exist */
+      _braid_BaseSumBasis(core, app, alpha, x->basis, beta, y->basis);
+   }
+   else if ( y->basis )
```
jbschroder, line 420: Just out of curiosity, do these cases of (x->basis = Null), but  (y->basis not Null) occur?

braid/base.c
```diff
@@ -466,6 +633,54 @@ _braid_BaseBufPack(braid_Core          core,
...
+       * header[-1] stores number of vectors,
+       * header[0] stores size of user vector 
+       * header[i] stores size of ith basis vector 
+       */
+      header = ((braid_Int*)buffer) + 1;
+      header[-1] = u->basis->rank + 1;                   /* total number of vectors */
+      head_size = (header[-1] + 1) * sizeof(braid_Int);  /* header[-1] + size of each vector */
+
+      data_buf = ((braid_Byte*)buffer) + head_size;      /* buffer must have type for legal pointer arithmetic */
+
```
jbschroder, line 658: Nice job packing all the vectors!

braid/base.h
```diff
@@ -78,6 +89,17 @@ _braid_BaseClone(braid_Core         core,     /**< braid_Core structure */
                  braid_BaseVector   u,        /**< vector to clone */ 
                  braid_BaseVector  *v_ptr     /**< output, newly allocated and cloned vector */ 
                  );
+
+/**
+ * This initializes a braid_Basis and calls the user's clone routine on each column.
```
jbschroder, line 94: Maybe say
"This initializes a braid_Basis by cloning A into B, using the user's clone routine"
?

braid/braid.c
```diff
@@ -65,6 +65,8 @@ braid_Drive(braid_Core  core)
       _braid_AdjointFeatureCheck(core);
    }
 
+   /* TODO: check for unsupported Delta features */
```
jbschroder, line 68: Do we want to add checks here, like init_basis must exist, if delta==True?

braid/braid.c
```diff
@@ -1840,3 +1896,82 @@ braid_SetTimings(braid_Core core,
...
+/*--------------------------------------------------------------------------
+ *--------------------------------------------------------------------------*/
+
+braid_Int
+braid_SetDeltaCorrection(braid_Core           core,      
+                         braid_Int            rank,      
+                         braid_PtFcnInitBasis init_basis,
+                         braid_PtFcnInnerProd inner_prod)
+{
+   /* TODO: compatibility checks */
```
jbschroder, line 1908: I might recommend most other compatibility checks to be done inside braid_Drive (that is, could remove this TODO, and just focus on the one in braid_Drive)

braid/braid.c
```diff
@@ -1840,3 +1896,82 @@ braid_SetTimings(braid_Core core,
...
+   if (_braid_CoreElt(core, richardson))
+   {
+      _braid_printf("  Braid: Delta correction is not currently compatible with Richardson error estimation, (feature coming soon?) disabling Richardson\n");
+      braid_SetRichardsonEstimation(core, 0, 0, 0);
+   }
+
+   if (_braid_CoreElt(core, adjoint))
+   {
+      _braid_printf("  Braid: Delta correction is not currently compatible with the adjoint feature, (coming soon?) not enabling Delta correction\n");
+      return _braid_error_flag;
```
jbschroder, line 1918: why return _braid_error_flag here, but not in other if statements?  maybe remove?

braid/braid.c
```diff
@@ -1840,3 +1896,82 @@ braid_SetTimings(braid_Core core,
...
+braid_Int
+braid_SetLyapunovEstimation(braid_Core core,
+                            braid_Int  relax,
+                            braid_Int  cglv,
+                            braid_Int  exponents
+                            )
+{
+   /**
+    * TODO: compatibility checks
+    * We could allow to turn this on even if Delta correction is off,
```
jbschroder, line 1962: Compatibility checks

braid/braid.h
```diff
@@ -147,6 +147,17 @@ typedef braid_Int
                    braid_Vector  *u_ptr          /**< output, newly allocated and initialized vector */
                    );
 
+/**
+ * (optional) Initializes a vector *u_ptr* at time *t*
```
jbschroder, line 151: "(optional) Initializes a delta correction basis vector *u_ptr* at time *t* "

braid/braid.h
```diff
@@ -190,19 +201,35 @@ typedef braid_Int
...
+                        );
+
 /**
  * Gives user access to XBraid and to the current vector *u* at time *t*.  Most
  * commonly, this lets the user write the vector to screen, file, etc...  The
  * user decides what is appropriate.  Note how you are told the time value *t*
  * of the vector *u* and other information in *status*.  This lets you tailor
  * the output, e.g., for only certain time values at certain XBraid iterations.
- * Querrying status for such information is done through
+ * Querying status for such information is done through
```
jbschroder, line 226: Thank you for catching all these typos!

braid/braid.h
```diff
@@ -1217,6 +1253,51 @@ braid_SetRichardsonEstimation(braid_Core core,                /**< braid_Core (_
                               braid_Int  local_order          /**< Local order of the time integration scheme, e.g., local _order=2 for backward Euler */
                               );
 
+/** 
+ * Turn on low-rank Delta correction. This uses Jacobians of the fine-grid time-stepper as a linear correction to the coarse time-stepper. 
+ * This can potentially greatly accelerate convergence for nonlinear systems.
+ * 
+ * The action of the Jacobian will be computed on a (low-rank) time-dependent basis initialized by the user.
+ * TODO: Finish this entry
+ */
```
jbschroder, line 1262: Is there any TODO left above?

braid/braid.h
```diff
@@ -1217,6 +1253,51 @@ braid_SetRichardsonEstimation(braid_Core core,                /**< braid_Core (_
...
+
+/**
+ * Defer the low-rank Delta correction to a coarse level or to a later iteration. 
+ * To mitigate some of the cost of Delta correction, it may be turned off on the first few fine-grids,
+ * or turned off for the first few iterations.
+ * 
+ */
+braid_Int
+braid_SetDeferDelta(braid_Core core,   /**< braid_Core (_braid_Core) struct*/
+                    braid_Int  level,  /**< Integer, Delta correction will be deferred to this level (Default 1)*/
```
jbschroder, line 1279: So, this skips delta correction on the fine level, and just starts it on the first coarse level, right?

braid/braid.pyx
```diff
@@ -86,7 +93,10 @@ cdef extern from "braid_status.h":
     ##
     # Wrap BufferStatus Routines
     int braid_BufferStatusGetMessageType (braid_BufferStatus status, int *messagetype_ptr)
+    int braid_BufferStatusGetTIndex ( braid_BufferStatus status, int *idx_ptr);
+    int braid_BufferStatusGetLevel ( braid_BufferStatus status, int *level_ptr);
     int braid_BufferStatusSetSize ( braid_BufferStatus status, int size);
+    int braid_BufferStatusSetBasisSize ( braid_BufferStatus status, int size);
```
jbschroder, line 99: Thank you David for catching lots of these small things missing, like this function wrappers

braid/braid_defs.h
```diff
@@ -48,6 +48,12 @@ typedef int    braid_Int;
 #define braid_Int_Max INT_MAX;
 #define braid_Int_Min INT_MIN;
 
+/**
+ * Defines byte type
+ * (can be any type, but sizeof(braid_Byte) MUST be 1)
+ */
+typedef char braid_Byte;
+
```
jbschroder, line 56: From what I can tell, a char is guaranteed to be 1 byte, 
https://www.ibm.com/docs/en/ibm-mq/7.5?topic=platforms-standard-data-types

braid/_braid.h
cadaverous-lives: It needs to be just as persistent as the tau correction (fa) since it is needed whenever tau correction is done. So it can be destroyed after interpolation, but it needs to persist until the end of interpolation, since the value at the current point is overwritten by the error correction, but we still need the basis vectors from the current point to step to the next point.. I'm thinking the simplest solution is to just copy the data I need to a temporary array, then use that in the step function instead of v[i-1]. That way, we are only storing a single set of basis vectors at a time, rather than one for every time-point owned by the processor.

braid/_braid.h
cadaverous-lives: This is here because it is only used in the context of StepStatus. This is how Lyapunov vectors are passed to the user's step function, and is only a reference, so there is no actual allocation for it either. I don't really have a problem with moving it up if you'd like.

braid/access.c
cadaverous-lives: Yes, this prevents having to add another argument to the user's step function that they will potentially never use.

braid/base.c
cadaverous-lives: This is a little over-engineered at the moment... I wrote it this way before I had finished the communication, and I wasn't sure that I would always be sending the basis in every communication, in which case the vector constructed by BufUnpack would have x->basis == Null. However, I think the way things are now this should never happen.

braid/base.c
cadaverous-lives: Thanks! I think it's pretty slick! I also like the typedef "braid_Byte". It has a nice ring to it :)

braid/braid.c
cadaverous-lives: I think that's a great idea. This is why I left a TODO in there. I was sure you guys would have better ideas for things to check for.

braid/braid.h
cadaverous-lives: I have an extension in my IDE that highlights them in red for me, so I can't miss them :)

braid/_braid.h
jbschroder: Oh, I see, maybe the comment could end with something like "..across the c-interval, for giving user access to vectors in Status routines"

I think that it's in the right place, since it deals with Status

braid/braid_status.c
```diff
@@ -224,6 +227,72 @@ braid_StatusGetCallingFunction(braid_Status status,
...
+   braid_Int cfactor = _braid_GridElt(grids[level], cfactor);
+
+   if (!_braid_StatusElt(status, lyap_exp) ||_braid_StatusElt(status, level) > 0 || (_braid_IsFPoint(index, cfactor) && nlevels > 1))
+   {
+      *num_returned = 0;
+      return _braid_error_flag;
+   }
+
+   braid_Int clower = _braid_GridElt(grids[0], clower);
+   braid_Int i_c    = (index - clower)/_braid_GridElt(grids[0], cfactor);
```
jbschroder, line 289: Can you assign "i_c" using this macro from braid.h?  If so, we should probably use the macro

+++++
#define _braid_MapFineToCoarse(findex, cfactor, cindex) \
( cindex = (findex)/(cfactor) )
++++++

braid/braid_status.c
```diff
@@ -729,6 +812,8 @@ _braid_StepStatusInit(braid_Real       tstart,
                       braid_Int        level,
                       braid_Int        nrefine,
                       braid_Int        gupper,
+                      braid_Int        calling_function,
```
jbschroder, line 815: Thank you for adding the calling function capability for StepStatus!

braid/braid_status.h
```diff
@@ -482,6 +510,16 @@ braid_StatusSetSize(braid_Status status,                   /**< structure contai
                     braid_Real   size                      /**< input, size of the send buffer */
                     );
 
+/**
```
jbschroder, line 513: Maybe change comment to  ?
"""
Set the size of the buffer for basis vectors.  If 
set by user, the send buffer will allocate "size" bytes 
of space for each basis vector.  If not, BufSize is used 
for the size of each basis vector.
"""

braid/braid_status.c
```diff
@@ -638,6 +718,9 @@ ACCESSOR_FUNCTION_GET4(Access, TILD,            Real, Int, Int, Int)
 ACCESSOR_FUNCTION_GET1(Access, WrapperTest,     Int)
 ACCESSOR_FUNCTION_GET1(Access, CallingFunction, Int)
 ACCESSOR_FUNCTION_GET1(Access, SingleErrorEstAccess,  Real)
+ACCESSOR_FUNCTION_GET1(Access, DeltaRank,       Int)
+ACCESSOR_FUNCTION_GET2(Access, LocalLyapExponents, Real, Int)
+ACCESSOR_FUNCTION_GET1_IN1(Access, BasisVec, Vector, Int)
```
jbschroder, line 723: Can the columns of text be aligned here, so that   Int/Real/Vector align?

braid/braid_status.c
```diff
@@ -761,6 +849,9 @@ ACCESSOR_FUNCTION_SET1(Step, RFactor,       Real)
 ACCESSOR_FUNCTION_SET1(Step, RSpace,        Real)
 ACCESSOR_FUNCTION_GET1(Step, Done,          Int)
 ACCESSOR_FUNCTION_GET1(Step, SingleErrorEstStep, Real)
+ACCESSOR_FUNCTION_GET1(Step, CallingFunction,    Int)
+ACCESSOR_FUNCTION_GET1(Step, DeltaRank,     Int)
+ACCESSOR_FUNCTION_GET1_IN1(Step, BasisVec, Vector, Int)
```
jbschroder, line 854: Can the columns of text be aligned here, so that   Real/Int/Real/Vector align?

braid/braid_status.h
```diff
@@ -644,6 +686,9 @@ ACCESSOR_HEADER_SET1(Step, RFactor,       Real)
 ACCESSOR_HEADER_SET1(Step, RSpace,        Real)
 ACCESSOR_HEADER_GET1(Step, Done,          Int)
 ACCESSOR_HEADER_GET1(Step, SingleErrorEstStep, Real)
+ACCESSOR_HEADER_GET1(Step, CallingFunction,    Int)
+ACCESSOR_HEADER_GET1(Step, DeltaRank, Int)
+ACCESSOR_HEADER_GET1_IN1(Step, BasisVec, Vector, Int)
```
jbschroder, line 691: Can the columns of text be aligned here, so that   Real/Int/Real/Vector align?

braid/communication.c
```diff
@@ -108,17 +108,16 @@ _braid_CommSendInit(braid_Core           core,
       /* Store the receiver rank in the status */
       _braid_StatusElt(bstatus, send_recv_rank) = proc;
 
-      /* Note that bufpack may return a size smaller than bufsize */ 
-      _braid_StatusElt(bstatus, size_buffer) = size;
```
jbschroder, line 112: Why is this line being removed?  Don't we want to tell the user what the current buffer size is through bstatus?

braid/delta.h
```diff
@@ -0,0 +1,75 @@
...
+#ifndef _braid_DELTA_HEADER
+#define _braid_DELTA_HEADER
+
+#include "_braid.h"
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
```
jbschroder, line 37: For consistency with the other header files, 
- There shouldn't be an indent
- The function description should be formatted like
/**
 * Function description
 * ...
 **/

braid/delta.h
```diff
@@ -0,0 +1,75 @@
...
+   braid_Int
+   _braid_LRDeltaDotMat(braid_Core core,
+                        braid_App app,
+                        braid_Basis psi,
+                        braid_Basis delta,
+                        braid_Basis basis);
+
+                     
+   /*----------------------------------------------------------------------------
+    * Perform modified Gram-Schmidt orthonormalization on a basis
```
jbschroder, line 62: while also computing exponents ?

braid/delta.c
```diff
@@ -0,0 +1,118 @@
...
+      _braid_CoreFcn(core, inner_prod)(app, basis->userVecs[i], u, &coords[i]);
+   }
+
+   /* compute the dot product of action with the new coordinate vector */
+   _braid_CoreFcn(core, sum)(app, coords[0], delta->userVecs[0], 0., u);
+   for (braid_Int i = 1; i < rank; i++)
+   {
+      _braid_CoreFcn(core, sum)(app, coords[i], delta->userVecs[i], 1., u);
+   }
+
```
jbschroder, line 52: I don't see a free for coords here.  

I would definitely do a valgrind with leakcheck=full on the code

braid/delta.c
```diff
@@ -0,0 +1,118 @@
...
+   for (braid_Int i = 0; i < rank; i++)
+   {
+      _braid_LRDeltaDot(core, app, psi->userVecs[i], delta, basis);
+   }
+
+   return _braid_error_flag;
+}
+
+braid_Int
+_braid_Normalize(braid_Core    core,
```
jbschroder, line 73: Is this function used anywhere?  Maybe I missed it.

braid/interp.c
```diff
@@ -76,31 +76,44 @@ _braid_FInterp(braid_Core  core,
...
+         {
+            /* can't overwrite va[fi-ilower], since it may be needed for
+             * Delta correction at the next time-point 
+             */
+            _braid_BaseClone(core, app, va[fi-ilower], &e);
+         }
+         else
+         {
+            e = va[fi-ilower];
+         }
```
jbschroder, line 98: Nice...I see now how you account for this over-writing of va

braid/relax.c
```diff
@@ -230,27 +230,46 @@ _braid_FCRelax(braid_Core  core,
...
-         
          }
          else if( (level == 0) && (ci == _braid_CoreElt(core, initiali)) && done)
          {
             /* If final opportunity for user access, provide access to initial condition */
             braid_BaseVector uref;
             _braid_UGetVectorRef(core, level, ci, &uref);
             _braid_AccessStatusInit(ta[ci-f_ilower], ci, rnm, iter, level, nrefine, gupper_zero,
-                                    done, 0, braid_ASCaller_FCRelax, astatus);
+                                    done, 0, braid_ASCaller_FCRelax, NULL, astatus);
```
jbschroder, line 272: Is there a reason why the basis argument is NULL?

braid/relax.c
```diff
@@ -216,7 +215,8 @@ _braid_FCRelax(braid_Core  core,
                /* Note that we initialize StepStatus here in a non-standard
                 * way, and hence cannot use _braid_Step(...). */
                _braid_StepStatusInit(time_left, ta[ci-ilower], ci-cfactor-1, tol,
-                                     iter, level, nrefine, gupper, status);
+                                     iter, level, nrefine, gupper, 
+                                     braid_ASCaller_FCRelax, NULL, status);
```
jbschroder, line 219: I'm just ignorant here -- can you explain why the basis argument is sometimes NULL (like here) and sometimes not NULL?

braid/residual.c
```diff
@@ -43,19 +44,45 @@ _braid_Residual(braid_Core        core,
...
 
+   /* this decides whether Delta correction needs to be computed on this level,
+   *  which depends on from where the residual is called
+   */
+   braid_Int delta_correct;
+   delta_correct = _braid_CoreElt(core, delta_correct)
+                   &&  ( iter  >= _braid_CoreElt(core, delta_defer_iter) )
+                   &&  ( level >= _braid_CoreElt(core, delta_defer_lvl) )
+                   && !( calling_function == braid_ASCaller_Residual
+                      && level == _braid_CoreElt(core, delta_defer_lvl) );
```
jbschroder, line 55: Why does this include both 
  level == _braid_CoreElt(core, delta_defer_lvl)
and 
  ( level >= _braid_CoreElt(core, delta_defer_lvl)
Is this a typo?

braid/residual.c
```diff
@@ -43,19 +44,45 @@ _braid_Residual(braid_Core        core,
...
+
+      /* By default: r = ustop - \Phi(ustart)*/
+      _braid_GetUInit(core, level, index, r, &rstop);
+      _braid_BaseStep(core, app, rstop, NULL, r, level, status);
+      _braid_CoreFcn(core, sum)(app, 1.0, ustop->userVector, -1.0, r->userVector);
+      _braid_BaseSumBasis(core, app, 0., r->basis, -1.0, r->basis);
+
+      return _braid_error_flag;
+   }
+   /* else, default behavior */
```
jbschroder, line 74: do we want a formal else statement here, and then a global return _braid_error_flag at the end of the routine?

braid/residual.c
```diff
@@ -79,24 +106,53 @@ _braid_FASResidual(braid_Core        core,
    _braid_Grid      **grids  = _braid_CoreElt(core, grids);
    braid_Int          ilower = _braid_GridElt(grids[level], ilower);
    braid_BaseVector  *fa     = _braid_GridElt(grids[level], fa);
+   braid_BaseVector  *va     = _braid_GridElt(grids[level], va);
+   // braid_Basis       *ba     = _braid_GridElt(grids[level], ba);
```
jbschroder, line 110: should this commented out line be removed?

braid/residual.c
```diff
@@ -79,24 +106,53 @@ _braid_FASResidual(braid_Core        core,
...
-   braid_Int        ii;
 
-   _braid_Residual(core, level, index, ustop, r);
-   if (level == 0)
+   braid_Int ii = index-ilower;
+
+   /* && short circuits, so (fa[ii] != NULL) is never evaluated if (level == 0) */
+   braid_Int delta_correct = _braid_CoreElt(core, delta_correct)
+                             && (_braid_CoreElt(core, niter) >= _braid_CoreElt(core, delta_defer_iter))
+                             && (level > _braid_CoreElt(core, delta_defer_lvl))
```
jbschroder, line 118: Should this be 

level > delta_defer_lvl

or

level >= delta_defer_lv

? 
I think that an above statement uses ">"

braid/residual.c
```diff
@@ -79,24 +106,53 @@ _braid_FASResidual(braid_Core        core,
...
+                             && (fa[ii] != NULL);
+
+   braid_BaseVector delta;    /* temporary storage for delta correction */
+
+   if ( delta_correct )
+   {  /* compute the Delta correction for the state and Lyapunov vectors */
+      _braid_BaseClone(core, app, r, &delta);
+      // _braid_LRDeltaDot(core, app, delta->userVector, fa[ii]->basis, ba[ii]);
+      // _braid_LRDeltaDotMat(core, app, delta->basis, fa[ii]->basis, ba[ii]);
```
jbschroder, line 127: can these two commented out lines be removed?

braid/restrict.c
```diff
@@ -95,6 +95,12 @@ _braid_FRestrict(braid_Core   core,
    braid_Real          *estimate;    
    braid_Real           factor, dtk, DTK;
 
+   /* Required for Delta correction */
+   braid_Int    delta_correct = (_braid_CoreElt(core, delta_correct) 
+                                 && iter >= _braid_CoreElt(core, delta_defer_iter) 
+                                 && level >= _braid_CoreElt(core, delta_defer_lvl) );
```
jbschroder, line 101: I'm seeing this logic now multiple places for ```delta_correct```

braid/step.c
```diff
@@ -40,21 +41,113 @@ _braid_Step(braid_Core         core,
...
-   _braid_StepStatusInit(ta[ii-1], ta[ii], index-1, tol, iter, level, nrefine, gupper, status);
+   braid_Int ii = index-ilower;
 
    /* If ustop is set to NULL, use a default approach for setting it */
    if (ustop == NULL)
    {
       _braid_GetUInit(core, level, index, u, &ustop);
    }
 
+   if (_braid_CoreElt(core, delta_correct) && iter >= _braid_CoreElt(core, delta_defer_iter) && level >= _braid_CoreElt(core, delta_defer_lvl))
```
jbschroder, line 58: Here's this logic again about whether delta_correct is enabled ... maybe a macro would make this more readable (and maintainable)

drivers/drive-ks-lib/ks_lib.cpp
```diff
@@ -0,0 +1,383 @@
+#include <limits>
```
jbschroder, line 1: Do we need the Braid Header up here?

drivers/drive-ks-lib/ks_lib.hpp
```diff
@@ -0,0 +1,71 @@
+#ifndef KS_LIB
```
jbschroder, line 1: Do we need the braid header up here?

drivers/drive-ks.cpp
```diff
@@ -0,0 +1,1412 @@
...
+}
+
+// Helper function to compute theta for a given level
+double
+MyBraidApp::getTheta(int level)
+{
+   return thetas[level];
+}
+
+void
```
jbschroder, line 324: The following 3 or 4 theta-themed functions don't have a comment describing what they do.  Could a little more commentary be added?

drivers/drive-lorenz-Delta-lib/lorenz_lib.cpp
```diff
@@ -0,0 +1,653 @@
+#include "lorenz_lib.hpp"
+
```
jbschroder, line 2: I know it's a driver, but maybe a short comment describing each function makes sense?  It's your driver, so take it as a suggestion

jbschroder:
Hi @cadaverous-lives 
This PR is a huge achievement.  The code is clean, nice, and does some pretty complicated stuff in a not-so-readable C-based library. 

One overall comment that I had on the drivers, is whether it makes sense to have two directories, one called something like
  ```KS_equation```
and the other
  ```Lorenz```

Then, all the lib, viz, and source files could go into those directories.

Otherwise, it's easy to miss the viz files, and the driver directory gets a bit cluttered.

drivers/drive-lorenz.c
```diff
@@ -362,7 +360,7 @@ int main (int argc, char *argv[])
    double        tstart, tstop;
    int           ntime;
 
-   int           max_levels = 1;
+   int           max_levels = 2;
```
jbschroder, line 363: Just curious...is this driver file redundant now?  No one really uses it, so it could get eliminated, assuming it's funcationality is now duplicated by example 7 (and the other Lorenz driver in this directory)

makefile.inc
```diff
@@ -119,9 +119,25 @@ else ifeq ($(findstring vulcan,$(HOSTNAME)),vulcan)
       CXXFLAGS = -O -Wall -fPIC
       FORTFLAGS = -O1 -Wall
    endif
+else ifeq ($(findstring cadaverous,$(HOSTNAME)),cadaverous)
```
jbschroder, line 122: Yes!  You do get your own line in the makefile.inc  :-)

test/lorenz_delta.saved
```diff
@@ -0,0 +1,239 @@
+# Begin Test 0
+
```
jbschroder, line 2: I got this difference
```
--- lorenz_delta.saved.0        2023-02-10 16:29:19.932943496 -0700
+++ std.out.0   2023-02-10 16:29:20.230988632 -0700
@@ -122,7 +120,7 @@
    braid_TestStepDiff:   v = B_0 - (u - v)/eps
    braid_TestStepDiff:  (v = step_dv(u)*A_0 - (step(u + eps*A_0) - step(u))/eps) 
    braid_TestStepDiff:   result = inner_prod(v, v) 
-   braid_TestStepDiff:   actual output:    result approx. 6.57e-19 
+   braid_TestStepDiff:   actual output:    result approx. 1.08e-18 
    braid_TestStepDiff:   expected output:  positive, near zero 
 Finished braid_TestStepDiff
 Starting braid_TestBufBasis
```
Not sure what to do...seems like numerical error though?

test/lorenz_delta.saved
```diff
@@ -0,0 +1,239 @@
...
+  residual norm         = 3.147082e-07
+  Average Lyapunov Exponents: [0.000000, 0.000000, 0.000000]
+
+# Begin Test 12
+
+  residual norm         = 8.077582e-07
+  Average Lyapunov Exponents: [0.008363]
+
+# Begin Test 13
```
jbschroder, line 210: I got about 5 tests with sometimes tiny differences like
```
--- lorenz_delta.saved.13       2023-02-10 16:29:19.933682822 -0700
+++ std.out.13  2023-02-10 16:29:22.172290355 -0700
@@ -1,5 +1,2 @@
-# Begin Test 13
-
-  residual norm         = 7.293948e-08
+  residual norm         = 7.293949e-08
   Average Lyapunov Exponents: [0.008363, -0.417301]
-
```

and sometimes larger differences like

```
jacobschroder$ more std.err.17
--- lorenz_delta.saved.17       2023-02-10 16:29:19.933880695 -0700
+++ std.out.17  2023-02-10 16:29:22.773064267 -0700
@@ -1,5 +1,2 @@
-# Begin Test 17
-
-  residual norm         = 5.637681e-14
+  residual norm         = 5.294091e-14
   Average Lyapunov Exponents: [0.000000, 0.000000, 0.000000]
-
jacobschroder$ more std.err.18
--- lorenz_delta.saved.18       2023-02-10 16:29:19.933933611 -0700
+++ std.out.18  2023-02-10 16:29:22.923200818 -0700
@@ -1,5 +1,2 @@
-# Begin Test 18
-
-  residual norm         = 5.181244e-14
+  residual norm         = 4.707584e-14
   Average Lyapunov Exponents: [-8.416132, -1.601542, -3.599683]
```

Overall, the regression tests look great!  

But maybe we should try limiting the number of XBraid iters to 2 or 3 in all the tests?  I think this will leave the resid norm larger (and less prone to loss of precision).  

Fewer time steps could also help, but if we drive the residual to close to 0, this may not help. 

Let's iterate a few times, trying first limiting the number of iterations, until my machine and yours give the same reg test numbers!

examples/ex-07.c
```diff
@@ -0,0 +1,772 @@
...
+ *                  {x' = sigma (y - x),    x(0) = -1.8430428,
+ *                  {y' = x (rho - z) - y,  y(0) = -0.07036326,
+ *                  {z' = xy - beta z,      z(0) = 23.15614636.
+ *
+ *                with sigma=10, beta=8/3, and rho=28,
+ *                and estimate the backward Lyapunov vectors along the trajectory
+ *                parallel-in-time using MGRIT equipped with Delta correction.
+ *                The initial condition should be a point close to the strange attractor.
+ *
+ *                TODO: Give some high-level info about the algorithm.
```
jbschroder, line 49: I would give a single high-level description of the algorithm in an .md file in /docs.  Then just reference that here.

examples/ex-07.c
```diff
@@ -0,0 +1,772 @@
...
+{
+   for (int i = 0; i < VecSize; i++)
+   {
+      VecAxpy(alpha, A[i], B[i]);
+   }
+}
+
+void MatVec(const MAT A, VEC x)
+{
+   VEC tmp;
```
jbschroder, line 121: is Vec tmp every deallocated?

doing a valgrind on some various parameter choices, with leakcheck=full, would probably be good

examples/ex-07.c
```diff
@@ -0,0 +1,772 @@
...
+
+} my_App;
+
+int my_App_Init(my_App *app,
+                MPI_Comm comm_,
+                double tstart_,
+                double tstop_,
+                int ntime_,
+                int rank_,
+                FILE *file,
```
jbschroder, line 208: can you add a comment on what file and file_lv point to?

examples/ex-07.c
```diff
@@ -0,0 +1,772 @@
...
+
+   double h; /* dt value */
+   h = tstop - tstart;
+
+   // get the number of Lyapunov vectors we need to propagate
+   int rank; /* rank of Delta correction */
+   braid_StepStatusGetDeltaRank(status, &rank);
+   MAT LinProp = {{0., 0., 0.}, {0., 0., 0.}, {0., 0., 0.}};
+
+   if (rank > 0) // we are propagating Lyapunov vectors
```
jbschroder, line 250: I would give a comment somewhere specifically spelling out what the user has to do in this function, e.g., something like

For delta correction, the user must propagate (i) the solution vector (as in traditional Braid) and (ii) the Lyapunov vectors.  The Lyapunov vectors are accessible through the status structure and are advanced with the Jacobian (see below)

Also, is the comment "we are propagating Lyapunov vectors" in the right spot?  It seems to come right before you propagate the solution vector.

examples/ex-07.c
```diff
@@ -0,0 +1,772 @@
...
+   u->values[0] = 9.614521163788712;
+   u->values[1] = 17.519127090401067;
+   u->values[2] = 13.94230712158098;
+
+   *u_ptr = u;
+
+   return 0;
+}
+
+int my_InitBasis(braid_App app, double t, int index, braid_Vector *u_ptr)
```
jbschroder, line 297: I would add a comment that this is a new function required for delta, and what you have to do with it

examples/ex-07.c
```diff
@@ -0,0 +1,772 @@
...
+{
+   for (int i = 0; i < VecSize; i++)
+   {
+      y->values[i] = alpha * x->values[i] + beta * y->values[i];
+   }
+
+   return 0;
+}
+
+int my_InnerProd(braid_App app, braid_Vector u, braid_Vector v, double *prod_ptr)
```
jbschroder, line 345: I would add a comment that this is a new function required for delta, and what you have to do with it

examples/ex-07.c
```diff
@@ -0,0 +1,772 @@
...
+
+   return 0;
+}
+
+int my_BufSize(braid_App app, int *size_ptr, braid_BufferStatus bstatus)
+{
+   *size_ptr = VecSize * sizeof(double);
+
+   /* tell braid the size of the basis vectors */
+   /* Note: this isn't necessary here, but for more complicated applications this size may be different */
```
jbschroder, line 424: I would again rephrase this in terms of something new, the user must do in comparison to traditional Braid, so say something like,

In contrast to traditional braid, you can also specify the size of a single Lyapunov basis vector (in case it is different from that of the state vector).
Note: this isn't necessary here, but for more complicated applications this size may be different

examples/ex-07.c
```diff
@@ -0,0 +1,772 @@
...
+   return 0;
+}
+
+int my_BufPack(braid_App app,
+               braid_Vector u,
+               void *buffer,
+               braid_BufferStatus bstatus)
+{
+   double *dbuffer = buffer;
+
```
jbschroder, line 435: Do you want to say,

This code is used to pack both the state vector and Lyapunov basis vectors.

?

examples/ex-07.c
```diff
@@ -0,0 +1,772 @@
...
+}
+
+int my_BufUnpack(braid_App app,
+                 void *buffer,
+                 braid_Vector *u_ptr,
+                 braid_BufferStatus status)
+{
+   double *dbuffer = buffer;
+   my_Vector *u;
+
```
jbschroder, line 452: Do you want to say,

This code is used to un-pack both the state vector and Lyapunov basis vectors.

?

test/lorenz_delta.sh
```diff
@@ -0,0 +1,165 @@
...
+        "$RunString -np 2 $example_dir/ex-07 -tstop 0.5 -ntime 256 -ml 3 -cf 2 -tol 1e-6 -rank 1 -defer-lvl 1" \
+        "$RunString -np 2 $example_dir/ex-07 -tstop 0.5 -ntime 256 -ml 3 -cf 2 -tol 1e-6 -rank 2 -defer-lvl 1" \
+        "$RunString -np 2 $example_dir/ex-07 -tstop 0.5 -ntime 256 -ml 3 -cf 2 -tol 1e-6 -rank 3 -defer-lvl 1" \
+        "$RunString -np 2 $example_dir/ex-07 -tstop 0.5 -ntime 256 -ml 3 -cf 2 -tol 1e-6 -rank 1 -defer-iter 1" \
+        "$RunString -np 2 $example_dir/ex-07 -tstop 0.5 -ntime 256 -ml 3 -cf 2 -tol 1e-6 -rank 2 -defer-iter 1" \
+        "$RunString -np 2 $example_dir/ex-07 -tstop 0.5 -ntime 256 -ml 3 -cf 2 -tol 1e-6 -rank 3 -defer-iter 1" \
+        "$RunString -np 2 $example_dir/ex-07 -tstop 0.5 -ntime 256 -ml 3 -cf 2 -tol 1e-6 -rank 1 -nu 1" \
+        "$RunString -np 2 $example_dir/ex-07 -tstop 0.5 -ntime 256 -ml 3 -cf 2 -tol 1e-6 -rank 2 -nu 1 -fcfLyap -noLyap" \
+        "$RunString -np 2 $example_dir/ex-07 -tstop 0.5 -ntime 256 -ml 3 -cf 2 -tol 1e-6 -rank 3 -noLyap" \
+        "$RunString -np 2 $example_dir/ex-07 -tstop 0.5 -ntime 256 -ml 3 -cf 2 -tol 1e-6 -rank 3 -noLyap -fcfLyap" )
```
jbschroder, line 107: it's often good to ahve one test that verifies the output is insensitive to numprocs ... it's not a must, though

examples/ex-07.c
```diff
@@ -0,0 +1,772 @@
...
+   /* Parse command line */
+
+   arg_index = 1;
+   while (arg_index < argc)
+   {
+      if (strcmp(argv[arg_index], "-help") == 0)
+      {
+         if (myid == 0)
+         {
+            printf("\n");
```
jbschroder, line 515:
```printf("Solve the Lorenz system with the following parameters\n");

File: examples/ex-07.c [hunk 61]
```diff
@@ -0,0 +1,772 @@
+/*BHEADER**********************************************************************
```
jbschroder, line 1: Can you add a short description to the examples/README.md file for this example?

braid/_braid.h
```diff
@@ -75,13 +75,26 @@ void _braid_ErrorHandler(const char *filename, braid_Int line, braid_Int ierr, c
...
+( (type *)realloc((braid_Byte *)ptr, (size_t)(sizeof(type) * (count))) )
 
 /** 
  * Free memory macro 
  **/
 #define _braid_TFree(ptr) \
-( free((char *)ptr), ptr = NULL )
+( free((braid_Byte *)ptr), ptr = NULL )
+
+/**
```
jbschroder, line 86: @rfalgout @cadaverous-lives Hi Rob, David found that he needed to allocate his braid_Basis struct with a little more syntatic sugar.  The old way of just using sizeof(braid_Basis) was only returning the size of a pointer, while he actually needs more space to store a couple data members (correct me if I'm wrong David).  

Anyway, this seems a reasonable change to me, but I thought we should discuss it.

rfalgout: I'm not sure we need the fancy Size0f macro.  Since all of this sizing is internal to braid, why not just use `sizeof(struct _braid_Vector_struct)` explicitly?  This macro assumes the `typedef` from opaque-object-name to pointer-to-struct follows a specific naming convention.  Of course, it currently does and it makes sense to continue that convention, but it's only a convention.

Take this with a grain of salt for now.  I will need to look more closely at what's happening here, because I don't fully understand it yet. :)  Stay tuned.

jbschroder: My guess is that since the rest of the code generally follows the convention of using the pointer (say braid_Vector) instead of the underlying struct (say _braid_Vector_struct), then David wanted to follow that convention with this macro.  

Maybe one option would be rename it 
_braid_SizeOfBraidType(...) 
so that any future developer would know this isn't a general SizeOf function, but a SizeOf function that assumes some kind of Braid conventions.  I'm not sure either!  I'll think on it some more :-)

test/lorenz_delta.saved
```diff
@@ -0,0 +1,121 @@
+# Begin Test 0
+
+   braid_TestSpatialNorm:   Test 1 Passed
```
jbschroder, line 3: Currently test 1 fails for me with this diff

    braid_TestBuf:   Test 1 Passed
 Finished braid_TestBuf: all tests passed successfully
    braid_TestInnerProd:   Test 1 Passed
-    braid_TestInitBasis:   Test 2 Passed
+    braid_TestInitBasis:   test 2 passed
    braid_TestBufBasis:   Test 1 Passed
    braid_TestBufBasis:   Test 2 Passed

braid/base.c
```diff
@@ -317,13 +411,48 @@ _braid_BaseSum(braid_Core        core,
...
    _braid_CoreFcn(core, sum)(app, alpha, x->userVector, beta, y->userVector);
-   _braid_CoreElt(core, timer_user_sum) += _braid_MPI_Wtime(core) - timer;
+   _braid_CoreElt(core, timer_user_sum) += _braid_MPI_Wtime(core, 2) - timer;
+
+   /* Sum over the basis vectors */
+   if ( x->basis && y->basis )
+   {  /* Add together the bases if they both exist */
+      _braid_BaseSumBasis(core, app, alpha, x->basis, beta, y->basis);
+   }
+   // else if ( y->basis )
```
jbschroder, line 423: @cadaverous-lives Can we remove these commented out lines?

braid/braid.c
```diff
@@ -58,12 +59,17 @@ braid_Drive(braid_Core  core)
...
    braid_Real     timer_drive_init;
 
-   timer_drive_init = _braid_MPI_Wtime(core);
+   timer_drive_init = _braid_MPI_Wtime(core, 2);
    /* Check for non-supported adjoint features */
    if (adjoint)
    {
       _braid_AdjointFeatureCheck(core);
    }
+   /* Check for non-supported Delta correction features */
```
jbschroder, line 68: @cadaverous-lives Take a look at these feature checks David.  See if I missed anything.

braid/relax.c
```diff
@@ -230,27 +230,46 @@ _braid_FCRelax(braid_Core  core,
                _braid_BaseFree(core, app,  bigstep);
             }
 
+            // /* Time-Average the computed Lyapunov exponents */
```
jbschroder, line 233: @cadaverous-lives Can we remove these commented out lines?

braid/residual.c
```diff
@@ -43,22 +44,47 @@ _braid_Residual(braid_Core        core,
...
+   {  /* Give the user access to the basis vectors through StepStatusGetBasisVec */
+      _braid_StepStatusInit(ta[ii-1], ta[ii], index-1, tol, iter, level, nrefine, gupper, calling_function, r->basis, status);
+
+      /* By default: r = ustop - \Phi(ustart) */
       _braid_GetUInit(core, level, index, r, &rstop);
-      _braid_BaseStep(core, app,  rstop, NULL, r, level, status);
-      _braid_BaseSum(core, app,  1.0, ustop, -1.0, r);
+      _braid_BaseStep(core, app, rstop, NULL, r, level, status);
+      _braid_CoreFcn(core, sum)(app, 1.0, ustop->userVector, -1.0, r->userVector);
+      _braid_BaseSumBasis(core, app, 0., r->basis, -1.0, r->basis);
```
jbschroder, line 69: @cadaverous-lives  I'm just curious why the constants are 0 and -1.  I guess you have to flip the sign for the basis in the residual?

drivers/KS-Delta/drive-ks.cpp
```diff
@@ -0,0 +1,1413 @@
+/*BHEADER**********************************************************************
```
jbschroder, line 1: @cadaverous-lives I would add a brief mention of the new drivers in the 
drivers/README.md 
file

jbschroder (approved):
@cadaverous-lives I left 4 or 5 really small comments.  After that, I'm good with a merge.

It also looks like there are two tiny conflicts to resolve.

jbschroder:
@rfalgout Hi Rob,  I just went through the most recent updates, and I think the PR is in good shape.  I just approved it, but there's a lot of code :-)  So, I could have definitely missed something.

braid/_braid.h
```diff
@@ -75,13 +75,13 @@ void _braid_ErrorHandler(const char *filename, braid_Int line, braid_Int ierr, c
  * Re-allocation macro 
  **/
 #define _braid_TReAlloc(ptr, type, count) \
-( (type *)realloc((char *)ptr, (size_t)(sizeof(type) * (count))) )
+( (type *)realloc((braid_Byte *)ptr, (size_t)(sizeof(type) * (count))) )
```
rfalgout, line 78: I still don't see the reason for `braid_Byte`.  Will this ever be anything other than `char`?  Is this useful to users?

braid/step.c
```diff
@@ -30,48 +30,142 @@ braid_Int
 _braid_Step(braid_Core         core,
             braid_Int          level,
             braid_Int          index,
+            braid_Int          calling_function,
```
rfalgout, line 33: Can you explain why this new `calling_function` argument is needed?  The `ASCaller` stuff is there to give the user some context about where they are in the MGRIT solver, since user code lives outside of braid.  The delta correction implementation in internal, so I am just wondering if there is another way to do this.  We might have to talk about this in person, but feel free to give it a shot here. :)

braid/base.c
```diff
@@ -92,7 +92,7 @@ _braid_BaseStep(braid_Core       core,
    {
       _braid_CoreElt(core, tnext) = _braid_CoreElt(core, tstop);
    }
-   timer = _braid_MPI_Wtime(core);
+   timer = _braid_MPI_Wtime(core, 2);
```
rfalgout, line 95: What's going on with this timing level thing?

jbschroder: Hi Rob, the timing level thing is something that I implemented.  Basically, the user can set two timing levels
1:  old behavior of Braid, where each iteration is timed and the whole solve in braid_Drive is timed.
2:  newer, more intrusive timing of user functions (primarily for TorchBraid).  When this is turned on, output files are created with detailed timing info for each MPI rank.

The "2" here passed to _braid_MPI_Wtime(...) tells braid to only call MPI_Wtime(..) if the timing level is at 2 or greater.   This way, we only call MPI_Wtime(...) when the user has asked for this more detailed information.  

When we did the timer PR a while back, I think we decided to wrap MPI_Wtime(), so that we only called it when needed.  This seemed reasonable -- but can always be improved :-)

rfalgout: Okay.  After I asked the question, I had a feeling this was not something David added, but that he was probably behind `master`?  If so, he should do a `git merge --no-ff master` into his branch before we finish the review and merge the PR.  I am happy with the timing level stuff.  It just didn't make sense to me in the context of this PR.  Thanks!

rfalgout (commented):
Hi @cadaverous-lives .  Nice work!  And lots of work, too!  I'd just like to discuss the few questions I asked and make sure you are current with `master` before giving the final go ahead to merge.  Let me know!

braid/base.c
jbschroder: Cool!  Sounds good.

cadaverous-lives:
@rfalgout, @jbschroder,
I think I have responded to all of your most recent comments, and I have made some small changes to the PR per your requests, as well as fixing the merge conflicts. Let me know if you need anything else!

braid/relax.c
jbschroder: @cadaverous-lives   Any thoughts on removing these commented out lines?

drivers/README.md
```diff
@@ -137,5 +130,14 @@ To run the examples, type
...
                 drivers/drive-adv-diff-1D-Cython/drive_adv_diff_1D-setup.py
 
-
-
+6. Directory drive-Lorenz-Delta/ implements the chaotic Lorenz system, with its trademark
+   butterfly shaped attractor. The driver uses the Delta correction feature and Lyapunov
+   estimation to solve for the backward Lyapunov vectors of the system and to accelerate
+   XBraid convergence. Visualize the solution and the Lyapunov vectors with *vis_lorenz_LRDelta.py*
+   Also see example 7 (examples/ex-07.c). *This driver is in a broken state, and needs*
+   *updating for compatibility with new Delta correction implementation.*
```
jbschroder, line 138: @cadaverous-lives Good to know about this driver.  We can remove this driver if you don't think it's worth fixing, or since it's a driver, it can probably be fixed after the PR.

braid/_braid.h
jbschroder: @cadaverous-lives What did you and Rob decide about braid_Byte?

braid/step.c
jbschroder: @cadaverous-lives @rfalgout Hi Guys, I was thinking about this too.  Is it that the calling function is needed to decide when to do certain parts of Delta Correction (internally to Braid), or does the user's Delta Correction code need calling function from Step to function properly?

cadaverous-lives:
@rfalgout, I guess the last 49 of my comments were trapped in limbo on github, but I was able to release them! I have been responding to your comments for the last 3 days, they just weren't showing up for you, until now. Sorry! :)

braid/_braid.h
rfalgout: It's coincidence that all of the other types have `braid_` names (it wasn't for style reasons).  The real reason for this is that we may need to change the underlying type to something different.  We need to do this in hypre to get around the fact that `int` is sometimes too small (for large problems, we need to deal with numbers bigger than 2 billion) or because we want to do arithmetic in a different precision.  We may need to do this in braid for similar reasons.

There is no reason for `braid_Byte` to ever be anything different than `char`.  However, I guess it doesn't hurt to leave it as is either, so let's do that.

braid/step.c
rfalgout: I'm not crazy about having it in the argument list of the key algorithmic routines (other than the status functions, where else is this currently used?).  However, this will require more discussion about how best to refactor everything and it's not worth the effort right now.  Since this is internal, it is easily changed later if a better approach presents itself.  Thanks!

rfalgout (approved):

braid/_braid.h
cadaverous-lives: I get that. Jacob and I were discussing this today and this is what we were thinking as well. I can always remove it after the pull request!

braid/step.c
cadaverous-lives: I took another look at _braid.h, and I see what you mean. The only functions in there now with calling_function in the arguments are _braid_Step and _braid_Residual... I'm not sure off the top of my head how I could refactor step to remove it, but I will think about it.

cadaverous-lives:
@rfalgout, @jbschroder,
Do you think this is ready to merge? I am not authorized to merge, so one of you will have to do it. Is there anything else?

rfalgout:
> @rfalgout, @jbschroder, Do you think this is ready to merge? I am not authorized to merge, so one of you will have to do it. Is there anything else?

If all of the tests are clean and such, yes.  I nominate Jacob to do the merge since he has paid much closer attention to many of the details. :)

rfalgout:
BTW, @jbschroder .  I have a preference for "squash and merge" now.  We've been doing that in hypre for quite some time.  I'd like to continue that practice here for braid if you don't mind.

jbschroder:
@rfalgout Sounds good -- I'll do a squash merge

PR #88, opened 2023-05-18
Linear MGRIT implementation

rfalgout:
This PR adds a linear MGRIT implementation to braid.  I made it a draft PR because we should change the "switch" to a runtime switch (add a `Set` function to turn it on/off) and we need to write some brief documentation and probably also a regression test.  Otherwise, the code has already been tested thoroughly on the `linear` branch (this `linear-pr` branch is the same, but is up to date with `master`)


rfalgout:
BTW, when I compile on my mac, I get the following warning:
```
Building ex-01-pp ...
mpicxx -g -Wall -fPIC -I../braid ex-01-pp.cpp -o ex-01-pp ../braid/libbraid.a -lm -lstdc++
ex-01-pp.cpp:292:4: warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]
   sprintf(filename, "%s.%04d.%03d", "ex-01.out", index, rank);
   ^
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/stdio.h:188:1: note: 'sprintf' has been explicitly marked deprecated here
__deprecated_msg("This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.")
^
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/sys/cdefs.h:215:48: note: expanded from macro '__deprecated_msg'
        #define __deprecated_msg(_msg) __attribute__((__deprecated__(_msg)))
                                                      ^
1 warning generated.
```
This has something to do with the code merged here from `master`.

jbschroder:
My homebrew clang 15.0.2 doesn't throw that warning, but from a little reading, some compilers (like Mac Developer Tools) have started throwing that warning.  Either way, snprintf seems to be C99 compliant, so switching to it shouldn't be a problem.

braid/restrict.c
```diff
@@ -222,6 +232,12 @@ _braid_FRestrict(braid_Core   core,
    /* Set initial guess on coarse level */
    _braid_InitGuess(core, c_level);
 
+   if (braid_LINEAR)
+   {
+      /* Don't need to update the rhs in the linear case (no Richardson either) */
+   }
+   else
+   {
```
jbschroder, line 240: Should we indent this code block?

braid/restrict.c
```diff
@@ -359,12 +376,19 @@ _braid_FRestrict(braid_Core   core,
       _braid_PrintSpatialNorms(core, tnorm_a, ncpoints);
    }
 
+   if (braid_LINEAR)
+   {
+      /* Don't need to update the rhs in the linear case (no Richardson either) */
+   }
+   else
+   {
```
jbschroder, line 384: Should we indent this code block?

PR #89, opened 2023-05-24
Julia interface

cadaverous-lives:
Introduces a new interface to XBraid, written in Julia, along with an example and a couple of drivers demonstrating the usage of the interface.

I originally did not intend to make a pull request for this, as I made it for my own personal use, so I have made some choices that may be too restrictive. I have made this a draft pr so we can discuss the choices made.

My goal was to write the interface so that the user does not have to manage any memory or communication, and as a result the user does not write clone, free, bufpack, or bufunpack routines, and some default functions are provided for sum, norm, and inner_prod, assuming the user's vector is a Julia array.

Julia also being a just-in-time compiled language resulted in the need for a "warmup" function, which allows the user's functions to be compiled before braid_Drive is called. Without this, if the first downcycle is skipped, the user's step function is compiled every time it is called on the coarse grid by a new processor, which adds a lot of extra serial work to the first iteration.

notable features missing:
- XBraid adjoint
- spatial coarsening/refinement
- residual option
- sync function

I believe all of these features are fairly straightforward to implement, I just haven't had a need to work on them.


braid/braid.c
```diff
@@ -509,24 +508,54 @@ braid_Destroy(braid_Core  core)
       _braid_TFree(_braid_CoreElt(core, tnorm_a));
       _braid_TFree(_braid_CoreElt(core, rdtvalues));
 
-      /* Destroy stored Lyapunov exponents */
-      if (_braid_CoreElt(core, lyap_exp))
+      if (grids[0] != NULL)
```
cadaverous-lives, line 511: This diff is a bit of a mess, but all I did was add this check for `grids[0] != NULL` to prevent a segfault in case the Braid hierarchy was never set up, which was happening occasionally while developing the Julia interface.

braid/braid.h
```diff
@@ -57,15 +57,15 @@ extern "C" {
  */
 
 /** Define Fortran name-mangling schema, there are four supported options, see braid_F90_iface.c */
-#define braid_FMANGLE 1
+#define braid_FMANGLE 0
 /** Turn on the optional user-defined spatial coarsening and refinement functions */ 
 #define braid_Fortran_SpatialCoarsen 0
 /** Turn on the optional user-defined residual function */ 
-#define braid_Fortran_Residual 1
```
cadaverous-lives, line 64: I had to set these all to zero in order to build libbraid.so. I don't really understand why, but the F90 interface seems to interfere with building the shared library. I would like to fix this!

braid/braid.jl/wrapper_functions.jl
```diff
@@ -0,0 +1,310 @@
...
+    ustop = unsafe_pointer_to_objref(_ustop)::BraidVector
+    status = StepStatus(_status)
+
+    tstart, tstop = status_GetTstartTstop(status)
+    delta_rank = status_GetDeltaRank(status)
+
+    # call the user's function
+    try
+        # residual option
+        if _fstop !== C_NULL
```
cadaverous-lives, line 72: Julia uses "multiple dispatch," i.e. function overloading, rather than null values to handle function arguments, so the user has to make sure their step function takes the right number and types of arguments.

braid/braid.h
jbschroder: I don't understand this either...I'm not sure if we want to turn off the FORTRAN interface, or leave instructions for julia users to do this

cadaverous-lives: Ok, I think I have this figured out. I guess clang doesn't automatically do a dynamic lookup for undefined symbols, so on my machine I need to add the flag -undefined dynamic_lookup. This fixes the error, so thankfully it seems it was just a 'me' problem.

cadaverous-lives: This is now fixed!

.gitignore
```diff
@@ -4,10 +4,12 @@
 *.out
 *.out.*
 *.csv
+*.gif
```
rfalgout, line 7: I wouldn't include '.gif' here

braid/braid.jl/wrapper_functions.jl
rfalgout: Makes sense to me

braid/braid_test.h
```diff
@@ -282,6 +308,32 @@ braid_TestDelta(braid_App               app,          /**< User defined App stru
                 braid_PtFcnStep         mystep        /**< Compute a time step with a braid_Vector */
                 );
 
+/**
+ * Warmup calls all user functions once with sensible arguments.
+ * This is especially useful for the Julia wrapper, since Julia is JIT compiled,
+ * and this force compiles all the functions used in braid_Drive.
+ */
+braid_Int
+braid_Warmup( braid_App                app,         /**< User defined App structure */
```
rfalgout, line 317: All the other functions in `braid_test` are for doing unit testing.  Does Warmup belong here or should the name be changed?

PR #90, opened 2023-05-24
New branch for julia interface pull request

cadaverous-lives:
(reopening on new branch)

Introduces a new interface to XBraid, written in Julia, along with an example and a couple of drivers demonstrating the usage of the interface.

I originally did not intend to make a pull request for this, as I made it for my own personal use, so I have made some choices that may be too restrictive. I have made this a draft pr so we can discuss the choices made.

My goal was to write the interface so that the user does not have to manage any memory or communication, and as a result the user does not write clone, free, bufpack, or bufunpack routines, and some default functions are provided for sum, norm, and inner_prod, assuming the user's vector is a Julia array.

Julia also being a just-in-time compiled language resulted in the need for a "warmup" function, which allows the user's functions to be compiled before braid_Drive is called. Without this, if the first downcycle is skipped, the user's step function is compiled every time it is called on the coarse grid by a new processor, which adds a lot of extra serial work to the first iteration.

notable features missing:

    XBraid adjoint
    spatial coarsening/refinement
    residual option
    sync function

I believe all of these features are fairly straightforward to implement, I just haven't had a need to work on them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ wrapperTests.txt
 # binaries
 # as these are build-dependent, binaries should not be committed
 *.o
+*.so
 *.a
 ex-01
 ex-01-adjoint
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,9 @@ all: braid examples
 
 braid: 
 	cd braid; $(MAKE)
+ifeq ($(shared),yes)
+		cd braid; $(MAKE) libbraid.so
+endif
 
 examples: ./braid/libbraid.a
 	cd examples; $(MAKE)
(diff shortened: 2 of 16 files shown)


cadaverous-lives:
I just merged a bunch of changes from my fork, including:
- support for sync function
- reorganized braid.jl directory
- added distinct status structure types (e.g. StepStatus, AccessStatus) so that now Julia knows which status functions can be called on each status object, and e.g. `XBraid.Status.getTStop(status::AccessStatus)` will result in a compile time error, rather than causing a potential segfault!

PR #92, opened 2023-06-01
Adds braid_StatusGetCFactor

cadaverous-lives:
braid_StatusGetCFactor(braid_Status status, braid_Int *cfactor_ptr, braid_Int level) for StepStatus and AccessStatus
- returns cfactor on given level
- along with small bugfix for braid_StatusGetBasisVec
I know in the past this has been achieved by storing the cfactor for each level in an array in braid_App, but I find this is annoying every time I have to do it, especially when this information is already stored in `core->grids`.
This is especially useful for my theta method implementation with adaptive refinement.

### DIFF
--- a/braid/braid_status.c
+++ b/braid/braid_status.c
@@ -115,6 +115,26 @@ braid_StatusGetLevel(braid_Status status,
 }
 
 braid_Int
+braid_StatusGetCFactor(braid_Status status,
+                       braid_Int   *cfactor_ptr,
+                       braid_Int    level
+                       )
+{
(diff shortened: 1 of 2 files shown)


rfalgout (approved):
Looks okay to me.

jbschroder:
@cadaverous-lives The return value for cfactor on the coarsest level is off, I think, in that returning a result there doesn't make much sense.  

I'm going to return -1, instead, for now, but we can of course change this.  I'm about ready to merge this.

jbschroder (approved):

PR #94, opened 2023-12-13
Function to query XBraid status abouth whether refinement will be accepted during s

cadaverous-lives:


cadaverous-lives:
This should probably include logic to check whether the `max_refine` limit has been reached, right?

PR #97, opened 2024-08-23
adding new feature that allows the user to turn off the global residu…

jbschroder:
@rfalgout  @eric-c-cyr

Hi Guys -- the TorchBraid folks have been asking for an easy way to turn off the global MPI Allreduce during the residual norm computation.  We usually have a fixed number of iterations, so this Allreduce is just extra work.  It turns out if we can turn off this Allreduce, it actually saves a decent amount of time in the GPU-MPI setting.

Take a look and let me know -- Thanks!


rfalgout (approved):
Looks good to me.  Thanks!

jbschroder:
Cool!  I will squash and merge then.

